PR #24, opened 2026-04-18
fix: use Coil for Palette color extraction

### LINKED ISSUE #6: Palette color extraction downloads full-size images without size limit
`PlayerChrome.kt` → `loadArtworkPresentation()` uses `URL(model).openStream()` to download artwork for Palette color extraction. This bypasses OkHttp (no timeout/cache config) and has no size constraint — if `artworkUri` points to an unscaled original, the full image is loaded into memory.

**Risk:** OOM on large album art, unnecessary bandwidth usage.

**Suggested fix:**
- Use OkHttp or Coil to load the image (respects timeouts, disk cache)
- Downsample before Palette analysis (e.g. `BitmapFactory.Options.inSampleSize`)
- Or use Coil's `ImageRequest` with a fixed size to get a pre-scaled bitmap

**Affected code:** `PlayerChrome.kt` lines ~490-500, `loadArtworkPresentation()` function.

PengDingkang:
Replace `URL.openStream()` with Coil `ImageRequest` for Palette color extraction in Now Playing.

**Before:** Raw `URL.openStream()` → no timeout, no cache, no size limit → OOM risk on large album art
**After:** Coil `ImageRequest` with `size(300)` → uses OkHttp (timeouts), disk/memory cache, downsampled to 300px

Closes #6

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/presentation/library/PlayerChrome.kt
+++ b/app/src/main/java/com/anzupop/saki/android/presentation/library/PlayerChrome.kt
@@ -1,6 +1,5 @@
 package com.anzupop.saki.android.presentation.library
 
-import android.graphics.BitmapFactory
 import androidx.activity.compose.BackHandler
 import androidx.compose.animation.AnimatedContent
 import androidx.compose.animation.AnimatedVisibility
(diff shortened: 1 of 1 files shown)


xiongnemo (approved):
Image

PR #33, opened 2026-04-18
feat: word-level karaoke lyrics with smooth fill

### LINKED ISSUE #29: Lyrics display raw inline timestamps for word-level synced LRC
## Bug

Some songs have word-level (per-character) synced lyrics where each character's timestamp is embedded inline in the `value` field:

```
寻[00:07.20]寻[00:08.62]觅[00:09.92]觅 [00:11.28]冷[00:12.44]冷[00:13.76]清[00:14.97]清[00:15.97]
```

The app displays these raw timestamps as part of the lyric text instead of stripping them.

## Cause

Navidrome embeds per-word timestamps in the `line.value` when the source LRC uses enhanced format (`[mm:ss.xx]` between characters). Our parser treats `value` as plain text.

## Phase 1: Strip inline timestamps (minimum fix)

Strip inline timestamps from `value` using regex before display:
```kotlin
val cleanText = value.replace(Regex("\\[\\d{2}:\\d{2}\\.\\d{2,3}]"), "")
```

This should be done in the mapper (`SubsonicMappers.toLyrics()`) when building `LyricLine.text`.

## Phase 2: Karaoke-style word-level highlight (future enhancement)

Parse the inline timestamps to extract per-character timing, then render with progressive highlight (like Apple Music / 网易云). Would need:
- Parse `value` into `List<WordCue(startMs, text)>`
- Custom `AnnotatedString` rendering with animated color per word based on `positionMs`
- Evaluate implementation difficulty when working on this

Song: 易安难安 by 赤羽

PengDingkang:
Word-level karaoke lyrics with smooth fill animation.

**Data layer:**
- `WordCue(startMs, text)` domain model for per-character timing
- Mapper parses inline `[mm:ss.xx]` timestamps from enhanced LRC format
- Strips raw timestamps from display text automatically

**Rendering:**
- Active line with word-level data: left-to-right `clipRect` fill animation
- `produceState` + `withFrameNanos` for 60fps position interpolation (global position updates at 500ms)
- Falls back to standard line-level highlight for non-word-level lyrics
- Zero overhead when lyrics hidden or paused

Closes #29

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/data/remote/subsonic/SubsonicMappers.kt
+++ b/app/src/main/java/com/anzupop/saki/android/data/remote/subsonic/SubsonicMappers.kt
@@ -227,10 +227,10 @@ internal fun SubsonicResponseDto.toLyrics(): SongLyrics? {
     return SongLyrics(
         synced = best.synced,
         lines = best.line.map { dto ->
-            val rawStart = (dto.start ?: 0L) - offset
-            val words = parseWordCues(rawStart.coerceAtLeast(0L), dto.value)
+            val rawStart = dto.start ?: 0L
+            val words = parseWordCues(rawStart, dto.value)
             LyricLine(
-                startMs = rawStart.coerceAtLeast(0L),
+                startMs = (rawStart - offset).coerceAtLeast(0L),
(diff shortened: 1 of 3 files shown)


xiongnemo (approved):
Image

PR #50, opened 2026-04-19
fix: show correct duration for transcoded streams

### LINKED ISSUE #42: Playback duration shows 00:00 when using transcoded (non-original) stream quality
## Bug

When streaming with a non-original quality (transcoded stream), the playback duration shows `00:00`. Original quality streams display duration correctly.

## Likely cause

Transcoded streams may not include duration metadata in the HTTP response headers or stream container. The player relies on the stream itself for duration info, but transcoded formats (e.g. MP3, Opus) served via HTTP may not have accurate length headers.

The original song metadata (from `getSong` or the album/search response) includes `durationSeconds`, but this may not be passed through to the MediaItem when building the stream request for transcoded playback.

## Expected behavior

Duration should always display correctly regardless of stream quality setting, using the metadata `durationSeconds` from the song data as fallback.

PengDingkang:
## Problem

Playback duration shows `00:00` when streaming with non-original quality (transcoded). Transcoded streams use chunked HTTP transfer without `Content-Length`, so ExoPlayer returns `C.TIME_UNSET`.

## Fix

When `player.duration` is unknown, fall back to `durationMs` from song metadata in MediaItem extras (from Subsonic API).

Closes #42

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/playback/DefaultPlaybackManager.kt
+++ b/app/src/main/java/com/anzupop/saki/android/playback/DefaultPlaybackManager.kt
@@ -349,7 +349,9 @@ class DefaultPlaybackManager @Inject constructor(
                 currentIndex = currentIndex,
                 queue = queue,
                 positionMs = player.currentPosition.coerceKnownTime(),
-                durationMs = player.duration.coerceKnownTime(),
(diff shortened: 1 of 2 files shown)


xiongnemo (approved):
Image

PR #60, opened 2026-04-20
feat: shuffle by virtual index mapping with seed-based persistence

### LINKED ISSUE #47: Optimize shuffle: reorder queue instead of using Media3 built-in shuffle
## Problem

When shuffle mode is enabled, several UX issues occur:

1. **Swipe direction inconsistent**: Left/right swipe in Now Playing still follows the original sequential order, not the shuffled order
2. **Cover art animation direction wrong**: Next/previous buttons cause the cover art to animate left or right inconsistently, because the shuffled track may be at any position in the original queue

Both issues stem from Media3's default shuffle implementation, which keeps the original queue order and only randomizes the "next track" selection internally. The queue indices and pager positions don't reflect the shuffled order.

## Proposed solution

When shuffle is toggled ON:
- Generate a shuffled copy of the current queue using a random seed
- Replace the media items in the player with the shuffled order
- Set the current track as the first item in the shuffled queue
- Disable Media3's built-in shuffle mode (since we handle it ourselves)

When shuffle is toggled OFF:
- Restore the original queue order
- Seek to the current track's position in the original order

This ensures:
- Swipe left/right matches next/previous track
- Cover art always animates in the correct direction
- Queue display in Now Playing shows the actual playback order

PengDingkang:
## Summary

Replace Media3's built-in shuffle with virtual index mapping for consistent UX and zero audio interruption.

## Approach

**Player queue stays untouched.** We maintain a `shuffleOrder: List<Int>` mapping (virtual index → player index) and translate all queue operations through it.

### Shuffle ON
1. Generate random seed via `System.nanoTime()`
2. Build deterministic shuffle order: `[currentTrack] + rest.shuffled(Random(seed))`
3. Store seed + anchor index to DataStore (12 bytes total)

### Shuffle OFF
1. Clear `shuffleOrder`
2. Clear DataStore shuffle state

### Auto-advance handling
Media3 auto-advances in original queue order. `onMediaItemTransition(REASON_AUTO)` intercepts this and redirects to the correct shuffled next track using `lastVirtualIndex`.

### Pager flash prevention
Queue updates are buffered via `stableQueue` — pager scrolls to correct page first, then queue snapshot updates. This prevents the cover art from briefly showing the wrong track.

### Persistence
- Seed (`Long`) + anchor index (`Int`) stored in DataStore
- On `restoreQueue`, rebuild identical shuffle order from saved seed
- Cleared on new `playQueue` / `playCachedQueue`

## What's translated through the mapping
- `skipToNext` / `skipToPrevious`
- `skipToQueueItem` / `removeQueueItem`
- `syncState` (queue order + currentIndex exposed to UI)

## Result
- Zero audio interruption on shuffle toggle ✅
- Swipe direction matches shuffled order ✅
- Cover art animates correctly ✅
- Queue view shows shuffled order ✅
- Shuffle state survives app restart ✅

Closes #47

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/data/repository/DataStorePlaybackPreferencesRepository.kt
+++ b/app/src/main/java/com/anzupop/saki/android/data/repository/DataStorePlaybackPreferencesRepository.kt
@@ -6,6 +6,7 @@ import androidx.datastore.preferences.core.booleanPreferencesKey
 import androidx.datastore.preferences.core.edit
 import androidx.datastore.preferences.core.emptyPreferences
 import androidx.datastore.preferences.core.intPreferencesKey
+import androidx.datastore.preferences.core.longPreferencesKey
 import androidx.datastore.preferences.core.stringPreferencesKey
 import com.anzupop.saki.android.domain.model.DEFAULT_STREAM_CACHE_SIZE_MB
 import com.anzupop.saki.android.domain.model.MAX_STREAM_CACHE_SIZE_MB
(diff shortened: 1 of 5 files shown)


xiongnemo (approved):
Image

PR #64, opened 2026-04-20
fix: non-blocking startup with first-reachable endpoint

### LINKED ISSUE #61: Startup blocks on endpoint probing, causing long loading spinner
## Problem

App startup sometimes shows a long loading spinner, likely blocking on endpoint probing. If all endpoints are unreachable (e.g. server down, network change), the user has to wait for every endpoint to timeout before seeing any UI.

## Expected behavior

- Show cached library content immediately (like playback session restore)
- Probe endpoints in the background, non-blocking
- UI becomes functional as soon as possible, endpoint status updates asynchronously
- If no endpoint is reachable, show the library from cache with an unobtrusive "offline" indicator

## Current behavior

Startup appears to wait for endpoint probing to complete before loading library content, causing a long spinner when endpoints are slow or unreachable.

PengDingkang:
Closes #61

## Problem

Startup blocks for a long time because:
1. `probe()` waits for **all** endpoints to finish (including unreachable ones timing out at 1.5s each)
2. `restorePlayQueue` fires network requests without waiting for probe, hitting unreachable endpoints with 10s connect timeout
3. `loadServerContent` network requests also don't benefit from probe results

### `probe()` — first-reachable-wins
- Returns as soon as the **first reachable endpoint** is found
- Remaining probes continue in background via `scope.launch` to find the optimal endpoint
- `probeVersion` updates incrementally so `sortedEndpoints` improves over time

### Startup sequence
1. **`loadCachedContent`** — immediately populates UI from `LibraryCacheRepository` (artists, albums, playlists, songs)
2. **`probe()`** — returns quickly once any endpoint responds
3. **`restorePlayQueue`** — now runs after probe, uses best known endpoint
4. **`refreshServerContent`** — network refresh with optimal endpoint order

### Network change reprobe
`onNetworkChanged` already calls `probe()` with exponential backoff delays (0s, 3s, 6s, 12s). The new `probe()` behavior means each round returns faster too.

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/data/remote/EndpointSelector.kt
+++ b/app/src/main/java/com/anzupop/saki/android/data/remote/EndpointSelector.kt
@@ -16,6 +16,7 @@ import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 import javax.inject.Singleton
 import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.SupervisorJob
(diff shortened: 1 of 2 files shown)


xiongnemo (approved):
Image

PR #69, opened 2026-04-20
fix: replace bufferedPosition heuristic with findCachedQualityKey

### LINKED ISSUE #63: isStreamCached false positive when seeking near end of track
## Bug

`isStreamCached` uses `bufferedPosition >= duration - 1000` to determine if a song is fully stream-cached. This is incorrect:

1. **Seek to end → false Cached**: Seeking near the end causes `bufferedPosition` to approach `duration`, triggering "Cached" label and tertiary color, even though the beginning/middle isn't cached
2. **Seek back → reverts to Streaming**: Dragging back resets `bufferedPosition`, so it flips back to "Streaming"
3. **Inconsistent with song list**: Song list uses `findCachedQualityKey` (real cache completeness check via `SimpleCache.isCached(key, 0, contentLength)`), but Now Playing uses the `bufferedPosition` heuristic — they disagree

## Fix

Replace `bufferedPosition` heuristic with `findCachedQualityKey` in `syncState`, guarded by `cacheVersion > 0` to avoid blocking main thread on first snapshot build.

PengDingkang:
Closes #63

## Problem

`isStreamCached` used `bufferedPosition >= duration - 1000` which gave false positives when seeking near end of track, and flipped back to "Streaming" when seeking backward.

## Fix

Replace the `bufferedPosition` heuristic with `findCachedQualityKey` — the same real cache completeness check used by the song list. This ensures Now Playing and song list agree on cached status.

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/playback/DefaultPlaybackManager.kt
+++ b/app/src/main/java/com/anzupop/saki/android/playback/DefaultPlaybackManager.kt
@@ -560,8 +560,9 @@ class DefaultPlaybackManager @Inject constructor(
 
         val currentRequest = player.currentMediaItem?.toPlaybackRequestOrNull()
         val streamCached = currentRequest != null && !currentRequest.isCached &&
-            player.bufferedPosition.coerceKnownTime() >= (player.duration.coerceKnownTime() - 1000L) &&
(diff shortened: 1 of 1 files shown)


xiongnemo (approved):
Image

PR #73, opened 2026-04-20
fix: suppress pager-driven skips after programmatic scroll

### LINKED ISSUE #72: Bug: tapping a song in Queue plays the previous track instead
## Bug

Tapping a song in the Queue bottom sheet frequently plays the wrong track, then the playback ping-pongs between two adjacent tracks.

## Root cause — confirmed via debug logs

Race condition between Queue tap and artwork pager's `snapshotFlow { settledPage }`:

```
23:26:44.717 queue tap: index=137 title=千本桜         ← user taps 137
23:26:44.736 skipToQueueItem: virtual=137               ← correct skip
23:26:44.764 pager settled: page=136 playbackIndex=137  ← pager hasn't caught up
23:26:44.784 skipToQueueItem: virtual=136               ← pager overrides to 136!
23:26:44.802 pager settled: page=137 playbackIndex=136  ← detects mismatch again
23:26:44.815 skipToQueueItem: virtual=137               ← ping-pong loop
```

The pager's `settledPage` lags behind the programmatic `animateScrollToPage`, causing it to fire `skipToQueueItem` with the old page index. This creates an infinite ping-pong between two tracks.

## Fix

Add a guard flag that suppresses pager-driven skips for a short window after a programmatic skip (queue tap, button skip). Only user-initiated pager swipes should trigger `skipToQueueItem`.

PengDingkang:
Closes #72

## Problem

Tapping a song in the Queue causes playback to ping-pong between two adjacent tracks. The artwork pager's `settledPage` lags behind `animateScrollToPage`, firing `skipToQueueItem` with the stale page index, creating an infinite loop.

## Fix

Add a 500ms suppression window after programmatic pager scrolls. During this window, `settledPage` changes are ignored, so only user-initiated swipes trigger track skips.

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/presentation/library/PlayerChrome.kt
+++ b/app/src/main/java/com/anzupop/saki/android/presentation/library/PlayerChrome.kt
@@ -328,16 +328,19 @@ fun NowPlayingOverlay(
         )
         // Sync pager when track changes externally (button skip, queue tap)
         var lastTrackId by remember { mutableStateOf(track.songId) }
+        // Guard: suppress pager-driven skips after programmatic scroll
+        var pagerSkipSuppressedUntil by remember { mutableStateOf(0L) }
         // Buffer queue to avoid pager flash during shuffle toggle:
         // update queue snapshot only after pager has scrolled to correct page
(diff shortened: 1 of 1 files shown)


xiongnemo (approved):
Image

PR #74, opened 2026-04-20
fix: resolve cover art URL at render time instead of queue build time

### LINKED ISSUE #62: Cover art fails to load in Now Playing
## Bug

Cover art in Now Playing sometimes fails to load, showing a blank/empty area. The song plays fine and metadata (title, artist) displays correctly, but the artwork is missing.

Observed on songs that have cover art visible in the song list. More likely to occur on poor network conditions.

## Root cause analysis

`buildArtworkUri` is called once at queue build time (`playSong`/`playQueue`), and the URL is baked into `PlaybackQueueItem.artworkUri`. It uses `candidates.firstOrNull()?.url` which picks the first endpoint from `endpointSelector.sortedEndpoints`.

Problems:
1. **Stale endpoint**: If the endpoint probe hasn't completed yet at queue build time, `sortedEndpoints` may return an unreachable endpoint → artwork URL points to a dead host
2. **No fallback**: Unlike API calls which use `executeWithFallback`, artwork loading goes through Coil with a single URL — no retry with alternate endpoints
3. **Silent failure**: `runCatching { ... }.getOrNull()` in `buildArtworkUri` swallows exceptions, returning null → no artwork

This explains why it's more common on poor networks — the first endpoint is more likely to be unreachable or slow.

## Possible fixes

1. **Use Coil interceptor with endpoint fallback**: If the first URL fails, retry with the next endpoint candidate
2. **Rebuild artwork URIs after endpoint probe completes**: When `endpointSelector` settles on the best endpoint, update the queue items' artwork URIs
3. **Pass `coverArtId` instead of pre-built URL**: Resolve the URL at render time using the current best endpoint, similar to how `resolveArtworkModel` works in the song list

PengDingkang:
Closes #62

## Problem

Cover art in Now Playing fails to load when the pre-built `artworkUri` points to an unreachable endpoint. The URL was baked into `PlaybackQueueItem` at queue build time using `candidates.firstOrNull()?.url`, which could pick an endpoint that hasn't been probed yet or has become unreachable.

## Fix

Replace pre-built `artworkUri` with render-time resolution via `resolveArtworkModel(server, coverArtId, null)`. This uses the current `ServerConfig` to build the cover art URL on each render, matching the pattern already used by song list artwork.

Changes:
- `queueArtworkModel()` now takes `ServerConfig?` and delegates to `resolveArtworkModel`
- Added `currentServer` param to `NowPlayingCapsule` and `QueueRow`
- All 5 call sites updated

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/presentation/SakiApp.kt
+++ b/app/src/main/java/com/anzupop/saki/android/presentation/SakiApp.kt
@@ -341,6 +341,7 @@ private fun RootShell(
             onSkipToQueueItem = onSkipToQueueItem,
             onRemoveQueueItem = onRemoveQueueItem,
             currentServer = uiState.servers.firstOrNull { it.id == track.serverId },
+            servers = uiState.servers,
             activeEndpointLabel = endpointStatus.activeEndpointLabel,
             activeEndpointId = endpointStatus.activeEndpointId,
             endpointProbeResults = endpointStatus.probeResults,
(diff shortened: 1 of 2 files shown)


xiongnemo (approved):
Image

PR #76, opened 2026-04-21
fix: remove serial buildArtworkUri calls from queue build

### LINKED ISSUE #75: Playback startup lag when playing from song list (serial buildArtworkUri calls)
## Bug

Playing a song from the song list / album / playlist has noticeable lag before playback starts. Playing from the Queue is instant.

## Root cause

`playQueue()` builds all MediaItems serially via `songs.map { }`. For each song it calls:
- `resolveQualityForSong()` → `findCachedQualityKey()` (cache snapshot lookup)
- `buildArtworkUri()` → `buildCoverArtRequest()` → `buildAuthenticatedCandidates()` (DB query for server config + endpoint sorting)

With 300+ songs this is 300 serial IO operations before playback can start.

## Fix

Since PR #74 resolves cover art URLs at render time, `buildArtworkUri` is no longer needed at queue build time. Pass `null` for `artworkUri` and remove `buildArtworkUri`. This eliminates the per-song network/DB overhead.

PengDingkang:
Closes #75

## Problem

Playing from song list / album / playlist has noticeable lag. `playQueue()` calls `buildArtworkUri()` serially for every song — each call does a DB query + endpoint sorting. With 300+ songs this blocks playback start.

## Fix

Remove `buildArtworkUri` calls entirely — cover art URLs are now resolved at render time (PR #74). Also removes the now-unused `SubsonicRepository` dependency from `DefaultPlaybackManager`.

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/playback/DefaultPlaybackManager.kt
+++ b/app/src/main/java/com/anzupop/saki/android/playback/DefaultPlaybackManager.kt
@@ -23,7 +23,6 @@ import com.anzupop.saki.android.domain.model.StreamQuality
 import com.anzupop.saki.android.domain.repository.PlaybackManager
 import com.anzupop.saki.android.domain.repository.PlaybackPreferencesRepository
 import com.anzupop.saki.android.domain.repository.StreamCacheRepository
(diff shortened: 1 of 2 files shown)


xiongnemo (approved):
Image

PR #79, opened 2026-04-21
fix: prevent bluetooth lyrics from corrupting next track title

### LINKED ISSUE #78: Bluetooth lyrics: song title corrupted when switching from lyrics to no-lyrics track
## Bug

When bluetooth lyrics is enabled and switching from a song with lyrics to a song without lyrics, the new song's title in the notification/lockscreen shows the previous song's title instead of its own.

## Root cause

`restoreOriginalTitle()` runs in the `finally` block when the lyrics coroutine is cancelled (on track change). By that time, `player.currentMediaItem` is already the **new** song. It overwrites the new song's correct title with the old song's `originalMediaTitle`.

## Fix

In `restoreOriginalTitle()`, skip restoration if the current media item has changed (compare `mediaId`). Or simply clear `originalMediaTitle` without restoring, since the new track's MediaItem already has the correct title.

PengDingkang:
Closes #78

## Problem

When bluetooth lyrics is enabled and switching from a song with lyrics to one without, the new song's notification title shows the previous song's title.

## Cause

`restoreOriginalTitle()` runs in the `finally` block when the lyrics coroutine is cancelled on track change. By that time `player.currentMediaItem` is already the new song, so it overwrites the new song's correct title with the old song's saved title.

## Fix

Track `originalMediaId` alongside `originalMediaTitle`. In `restoreOriginalTitle()`, skip restoration if `currentMediaItem.mediaId` doesn't match the saved ID.

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/playback/SakiPlaybackService.kt
+++ b/app/src/main/java/com/anzupop/saki/android/playback/SakiPlaybackService.kt
@@ -76,6 +76,7 @@ class SakiPlaybackService : MediaSessionService() {
     private var player: ExoPlayer? = null
     private var mediaSession: MediaSession? = null
     private var originalMediaTitle: CharSequence? = null
+    private var originalMediaId: String? = null
     private var soundBalancingMode = SoundBalancingMode.OFF
     private var loudnessEnhancer: LoudnessEnhancer? = null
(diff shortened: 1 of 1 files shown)


xiongnemo (approved):
Image

PR #84, opened 2026-04-21
feat: manual endpoint selection in Now Playing status dialog

### LINKED ISSUE #82: Manual endpoint selection in Now Playing endpoint status dialog
## Feature

Allow users to manually select an endpoint from the Now Playing endpoint status dialog. Reachable endpoints are clickable, unreachable ones are disabled/grayed out.

## Behavior

- Tap a reachable endpoint card → force that endpoint as active
- Force persists until next network change (`onNetworkChanged` clears it)
- If the forced endpoint fails during an actual request (`invalidate()`), the force is cleared and normal fallback resumes
- Probe continues running normally to update latency/reachability UI, but does not override the forced selection

## Implementation

- Add `forcedEndpoints` map + `forceEndpoint()` to `EndpointSelector`
- Guard `probe()` to not overwrite `bestEndpoints` when forced
- `onNetworkChanged()` clears forced
- `invalidate()` clears forced
- Wire callback through ViewModel → PlayerChrome endpoint status dialog

PengDingkang:
Closes #82 (replaces #83 which was merged into a stale branch)

## Feature

Tap a reachable endpoint in the Now Playing endpoint status dialog to force it as active. Tap again to cancel and return to automatic selection.

## Behavior

- Reachable endpoints are clickable, unreachable ones are grayed out / disabled
- Forced endpoint shows 📌 indicator
- Force persists until:
  - Network change (`onNetworkChanged` clears it)
  - Request failure on that endpoint (`invalidate` clears it, normal fallback resumes)
- Probe continues running to update latency/reachability UI but does not override forced selection
- Forced endpoint stays clickable even when unreachable so user can always unpin

## Changes

- `EndpointSelector`: `forcedEndpoints` map, `forceEndpoint()`, `clearForce()`, `isForced()`, guarded `probe()`/`invalidate()`/`onNetworkChanged()`/`unregisterServer()`
- `SakiAppViewModel`: toggle logic in `forceEndpoint()`, `isForced` in `EndpointStatus`
- `PlayerChrome`: `Surface(onClick)` with `enabled = result.reachable || (isActive && isEndpointForced)`, disabled styling

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/data/remote/EndpointSelector.kt
+++ b/app/src/main/java/com/anzupop/saki/android/data/remote/EndpointSelector.kt
@@ -44,6 +44,7 @@ class EndpointSelector @Inject constructor(
     private val scope = CoroutineScope(SupervisorJob() + ioDispatcher)
 
     private val bestEndpoints = ConcurrentHashMap<Long, Long>()
+    private val forcedEndpoints = ConcurrentHashMap<Long, Long>()
     private val lastProbeResults = ConcurrentHashMap<Long, List<EndpointProbeResult>>()
     private val serverConfigs = ConcurrentHashMap<Long, ServerConfig>()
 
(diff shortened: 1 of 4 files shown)


xiongnemo (approved):
Image

PR #90, opened 2026-04-21
fix: use active endpoint for cover art URL after network switch

### LINKED ISSUE #87: Cover art fails to load after network switch
## Bug

After switching networks (e.g. cellular → WiFi), cover art fails to load. The artwork URL points to the old endpoint which is no longer reachable.

## Root cause

`resolveArtworkModel()` / `buildCoverArtUrl()` picks the first endpoint from the server config and builds a static URL. After a network switch, `EndpointSelector` re-probes and selects a different best endpoint, but:

1. **UI side**: Coil has cached the old URL key — no recompose triggers a new URL
2. **Notification/lockscreen**: `artworkUri` baked into MediaItem metadata by `resolvePlayableItem` (PR #76) is never updated after network change

### UI side
Make `probeVersion` a key in artwork resolution so that when the active endpoint changes, Coil gets a new URL and reloads.

### Notification side
On `onMediaItemTransition` or network change, re-resolve the current item's artwork metadata and update via `replaceMediaItem`.

PengDingkang:
Closes #87

## Problem

Cover art fails to load after network switch. Two issues:

1. **Wrong endpoint**: `buildCoverArtUrl` generates a URL with the first endpoint by order, which may be unreachable after network change
2. **Cache duplication**: If the URL changes per endpoint, Coil caches the same image multiple times (once per endpoint)

## Solution

Two-layer approach — stable cache key + dynamic endpoint at request time:

**Artwork.kt** — Generates a stable URL using the first endpoint + deterministic salt (derived from coverArtId, not random). This URL never changes for the same image, so Coil's disk/memory cache works correctly with a single copy.

**CoverArtEndpointInterceptor** (OkHttp) — On cache miss, when Coil actually makes an HTTP request, this interceptor rewrites the URL's scheme+host+port to the current best endpoint from `EndpointSelector`. On cache hit, no network request is made and the interceptor is never called.

**EndpointSelector.findServerByHostPort** — Looks up which server a request belongs to by matching the URL's host+port against registered endpoints.

## Result

- Same image cached once regardless of endpoint changes ✅
- Network switch → new images load via best endpoint ✅  
- Cached images load instantly without network ✅
- No UI component changes needed ✅

## Files changed

- `CoverArtEndpointInterceptor.kt` — New OkHttp interceptor, rewrites cover art requests to best endpoint
- `EndpointSelector.kt` — Added `findServerByHostPort()`
- `NetworkModule.kt` — Register interceptor with `dagger.Lazy` to avoid dependency cycle
- `Artwork.kt` — Added doc comment on `buildCoverArtUrl` explaining stable URL strategy"

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/data/remote/CoverArtEndpointInterceptor.kt
+++ b/app/src/main/java/com/anzupop/saki/android/data/remote/CoverArtEndpointInterceptor.kt
@@ -1,0 +1,71 @@
+package com.anzupop.saki.android.data.remote
+
+import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
+import okhttp3.Interceptor
+import okhttp3.Response
+
+/**
+ * OkHttp interceptor that rewrites Coil cover art request URLs to use the best available endpoint.
+ *
+ * Coil generates cover art URLs with a fixed base (first endpoint by order) for stable caching.
+ * When the actual HTTP request is made (cache miss), this interceptor replaces the base URL
+ * with the current best endpoint so the request reaches a reachable server.
+ *
+ * Only rewrites requests whose host+port matches the canonical (first-by-order) endpoint,
+ * so repository-generated multi-endpoint fallback requests are not affected.
+ */
+class CoverArtEndpointInterceptor(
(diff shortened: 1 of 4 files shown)


xiongnemo (approved):
Image

PR #92, opened 2026-04-21
fix: truncate file before writing config export

### LINKED ISSUE #91: Config export produces corrupted JSON when overwriting existing file
## Bug

Exporting config to an existing file produces corrupted JSON. The new JSON is written but the file is not truncated first, so if the new content is shorter than the old, leftover bytes from the previous export remain at the end.

## Root cause

`contentResolver.openOutputStream(uri)` defaults to `"w"` mode which does not truncate SAF-backed files. When the user exports to the same filename, the old content is only partially overwritten.

## Fix

Use `openOutputStream(uri, "wt")` to truncate before writing.

PengDingkang:
Closes #91

`openOutputStream(uri)` → `openOutputStream(uri, "wt")` to truncate the file before writing. Without this, exporting to an existing file leaves leftover bytes when the new JSON is shorter than the old.


xiongnemo (approved):
LGTM

PR #94, opened 2026-04-21
fix: re-probe endpoints when server config changes

### LINKED ISSUE #93: Endpoint changes don't take effect until manual reprobe
## Bug

Adding or removing endpoints in server settings does not take effect until the user manually triggers a reprobe from the Now Playing endpoint status dialog.

## Root cause

`handleServerConfigsChanged` only calls `endpointSelector.probe()` when the selected server ID changes or on first load. Editing endpoints on the same server doesn't change the server ID, so the new endpoints are never registered with `EndpointSelector`.

## Fix

Re-register and re-probe the selected server whenever its config changes, even if the server ID stays the same.

PengDingkang:
Closes #93

When the selected server's config changes (endpoints added/removed/edited) without a server ID change, re-register and re-probe the server with `EndpointSelector` so new endpoints take effect immediately.


xiongnemo (approved):
LGTM

PR #96, opened 2026-04-22
fix: adaptive quality shuffle + cover art crash on network switch

### LINKED ISSUE #86: Adaptive quality: stream quality jumps inconsistently on network switch
## Bug

When switching from cellular to WiFi with adaptive quality enabled, the stream quality jumps inconsistently: e.g. 320kbps → original → 320kbps → 320kbps (all 320 after that). The first track after the switch gets the wrong quality.

## Likely cause

Race condition between `rebuildUpcomingItems()` (triggered by `networkType` change) and `resolvePlayableItem()` (triggered by preload/playback). When the network changes:

1. `rebuildUpcomingItems()` replaces upcoming MediaItem metadata with the new quality
2. But if preload has already started resolving the next item via `resolvePlayableItem()` using the old network type, that item keeps the old quality
3. The replaced MediaItem triggers a new `onAddMediaItems` → `resolvePlayableItem`, but the timing of `networkType` StateFlow propagation vs the resolve call is non-deterministic

## Reproduction

1. Enable adaptive quality (WiFi: original, cellular: 320kbps)
2. Start playing on cellular
3. Switch to WiFi mid-playback
4. Observe: next track briefly plays at original, then subsequent tracks revert to 320kbps (or vice versa)

### LINKED ISSUE #95: Crash: CoverArtEndpointInterceptor throws IllegalStateException on network switch
## Bug

App crashes with `IllegalStateException: state: 0` in OkHttp's `Http1ExchangeCodec.readResponseHeaders` when switching networks (WiFi ↔ cellular) while cover art is loading.

## Stack trace

```
FATAL EXCEPTION: OkHttp Dispatcher
java.lang.IllegalStateException: state: 0
  at okhttp3.internal.http1.Http1ExchangeCodec.readResponseHeaders(Http1ExchangeCodec.kt:173)
  at okhttp3.internal.connection.Exchange.readResponseHeaders(Exchange.kt:110)
  at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.kt:93)
  ...
  at CoverArtEndpointInterceptor.intercept(CoverArtEndpointInterceptor.kt:26)
  ...
```

## Cause

Network switch kills the underlying TCP connection, but OkHttp's connection pool still holds the stale connection. `CoverArtEndpointInterceptor` rewrites the URL to the best endpoint and calls `chain.proceed()`, which uses the dead connection. OkHttp throws `IllegalStateException` (not `IOException`), so the retry interceptor doesn't catch it.

## Fix direction

Wrap `chain.proceed()` in the interceptor with a try-catch for `IllegalStateException`, and either retry or let it fall through as an `IOException` so OkHttp's retry logic handles it.

## Reproduction

1. Start playing with cover art visible
2. Switch network (WiFi → cellular or vice versa)
3. App crashes within a few seconds

Found while debugging #86.

PengDingkang:
Closes #86, fixes #95

## Problem

Adaptive quality doesn't switch correctly in shuffle mode after network change. Two root causes:

### 1. Wrong rebuild indices (#86)

`rebuildUpcomingItems()` iterated physical indices `current+1..+10`, which in shuffle mode are **not** the next tracks to play. The actual next tracks (determined by `SakiShuffleOrder`) were never rebuilt.

### 2. `replaceMediaItem` corrupts ShuffleOrder (#86)

`replaceMediaItem()` internally does remove+insert, which calls `ShuffleOrder.cloneAndRemove().cloneAndInsert()`. Each replaced item gets appended to the **end** of the shuffle order. After 10 replacements, ExoPlayer's shuffle order is completely different from ours, so "next track" plays an un-rebuilt item.

### 3. CoverArtEndpointInterceptor crash (#95)

Network switch kills TCP connections, but OkHttp's connection pool holds stale ones. The interceptor's `chain.proceed()` throws `IllegalStateException` (not `IOException`), bypassing OkHttp's retry logic and crashing the app.

## Fix

- `rebuildUpcomingItems()` now follows `shuffleDisplayOrder` to find the correct next 10 tracks in shuffle mode
- After all `replaceMediaItem` calls, re-sends the deterministic `ShuffleOrder` to the service to restore the correct playback order
- `CoverArtEndpointInterceptor` wraps the rewritten request's `chain.proceed()` in try-catch, converting `IllegalStateException` to `IOException`

## Tested

- Shuffle + cellular → WiFi: next track plays at Original quality ✅
- Shuffle + WiFi → cellular: next track plays at 320kbps ✅
- Sequential playback: unchanged behavior ✅
- Network switch no longer crashes the app ✅

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/data/remote/CoverArtEndpointInterceptor.kt
+++ b/app/src/main/java/com/anzupop/saki/android/data/remote/CoverArtEndpointInterceptor.kt
@@ -1,5 +1,6 @@
 package com.anzupop.saki.android.data.remote
 
+import java.io.IOException
 import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import okhttp3.Interceptor
 import okhttp3.Response
@@ -66,6 +67,11 @@ class CoverArtEndpointInterceptor(
             .encodedPath(bestBasePath + relativePath)
             .build()
 
-        return chain.proceed(request.newBuilder().url(newUrl).build())
+        return try {
+            chain.proceed(request.newBuilder().url(newUrl).build())
+        } catch (e: IllegalStateException) {
+            // Stale connection after network switch — rethrow as IOException so OkHttp retries
+            throw IOException("Stale connection after endpoint rewrite", e)
+        }
     }
 }
--- a/app/src/main/java/com/anzupop/saki/android/data/remote/CoverArtEndpointInterceptor.kt
+++ b/app/src/main/java/com/anzupop/saki/android/data/remote/CoverArtEndpointInterceptor.kt
@@ -30,33 +30,33 @@ class CoverArtEndpointInterceptor(
         val endpointSelector = endpointSelectorProvider()
 
         val matchedServerId = endpointSelector.findServerByHostPort(url.host, url.port)
-            ?: return chain.proceed(request)
+            ?: return proceedSafely(chain, request)
 
         // Only rewrite if this request uses the canonical (first-by-order) endpoint.
         // Non-canonical requests come from SubsonicRepository fallback and should not be rewritten.
         val canonicalEndpoint = endpointSelector.getCanonicalEndpoint(matchedServerId)
-            ?: return chain.proceed(request)
+            ?: return proceedSafely(chain, request)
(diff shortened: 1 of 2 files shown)


xiongnemo (approved):
Image

PR #97, opened 2026-04-22
feat: resolve stream URLs at play time via ResolvingDataSource

### LINKED ISSUE #86: Adaptive quality: stream quality jumps inconsistently on network switch

### LINKED ISSUE #98: Investigation: adaptive quality architecture and ResolvingDataSource solution
## Summary

Adaptive quality (switching stream quality based on WiFi/cellular) had a fundamental architectural issue: stream URLs were resolved at queue build time, but needed to change at play time when the network switches. Multiple fix attempts were needed to find the correct solution.

### Original bug (#86)

When switching networks with adaptive quality enabled and shuffle on, the next track plays at the wrong quality. E.g., WiFi→cellular should switch from Original to 320kbps, but tracks keep playing at Original.

### Investigation

Debug logging revealed the issue had **two layers**:

1. **Wrong rebuild indices**: `rebuildUpcomingItems()` used physical indices `current+1..+10` which in shuffle mode are not the next tracks to play. The actual next tracks (determined by `SakiShuffleOrder`) were never rebuilt.

2. **`replaceMediaItem` corrupts ShuffleOrder**: Even after fixing the indices, `replaceMediaItem()` internally does remove+insert, which calls `ShuffleOrder.cloneAndRemove().cloneAndInsert()`. Each replaced item gets appended to the **end** of the shuffle order. After 10 replacements, ExoPlayer's shuffle order is completely different from ours.

### First fix attempt (#96)

- Fixed shuffle indices in `rebuildUpcomingItems`
- Added `sendShuffleOrderToService` after replacements to restore ShuffleOrder
- Also fixed CoverArtEndpointInterceptor crash (#95)

**Result**: ShuffleOrder resync via async custom command was unreliable — the 10 `replaceMediaItem` calls trigger timeline changes that race with the ShuffleOrder command. ExoPlayer's shuffle order was still corrupted.

### Root cause

The problem was **architectural**: resolving stream URLs at queue build time (in `onAddMediaItems`) and trying to update them later via playlist manipulation (`replaceMediaItem`) is fundamentally incompatible with ExoPlayer's internal state management (ShuffleOrder, preload, timeline).

### Final fix (#97)

Moved URL resolution from queue build time to **play time** using Media3's `ResolvingDataSource`:

- `resolvePlayableItem` builds placeholder URIs (`saki://stream?serverId=X&songId=Y`)
- `ResolvingDataSource.Resolver` resolves these to real Subsonic stream URLs when ExoPlayer actually opens the data source
- Quality is determined by the **current** network type at the moment of playback
- No playlist manipulation needed → ShuffleOrder never touched → shuffle works correctly

### DataSource pipeline

```
ResolvingDataSource (resolves placeholder → real URL + cache key)
  → CacheDataSource (cache lookup with resolved key)
    → OkHttp (on cache miss)
```

## Key lessons

1. **Fix at the right abstraction layer**: Quality is a data stream property, not a playlist property. It should be resolved when the stream opens, not when the playlist is built.

2. **`replaceMediaItem` has hidden side effects in shuffle mode**: It internally does remove+insert which rebuilds ShuffleOrder. This is not documented and makes it unsuitable for updating items in a shuffled playlist.

3. **Debug logging was essential**: Each iteration of logging revealed a different layer of the problem — wrong indices, ShuffleOrder corruption, async timing issues.

## Related

- #86 — Original bug report
- #95 — CoverArtEndpointInterceptor crash on network switch (fixed in #96)
- #96 — First fix attempt (shuffle index + ShuffleOrder resync)
- #97 — Final fix (ResolvingDataSource)

PengDingkang:
Closes #86

## Problem

Adaptive quality doesn't switch correctly after network change, especially in shuffle mode. See #86 and the investigation in #97 for full details.

### Root cause

Stream URLs (with baked-in `maxBitRate` parameter) were resolved once at queue build time in `onAddMediaItems`. After a network switch, the only way to update them was `replaceMediaItem`, which internally does remove+insert and **corrupts ExoPlayer's ShuffleOrder** in shuffle mode. The previous fix (#96) attempted to resync ShuffleOrder after replacement, but the async IPC timing made this unreliable.

## Solution

Move URL resolution from queue build time to **play time** using Media3's `ResolvingDataSource`.

### How it works

1. `resolvePlayableItem` now builds a **placeholder URI** (`saki://stream?serverId=1&songId=abc`) instead of a real stream URL
2. `ResolvingDataSource.Resolver` intercepts these placeholders when ExoPlayer actually opens the data source, resolving them to real Subsonic stream URLs based on the **current** network type
3. `CacheDataSource` sits inside `ResolvingDataSource`, so cache lookups use the resolved URL and cache key

### Pipeline

```
ResolvingDataSource (placeholder → real URL + cache key)
  → CacheDataSource (cache lookup with correct key)
    → OkHttp (on cache miss)
```

### What's removed

- `rebuildUpcomingItems()` — no longer needed, quality is resolved at play time
- `networkType` collector in `DefaultPlaybackManager.init` — no longer needed
- All `replaceMediaItem` calls for quality updates — no playlist manipulation needed

### UI quality label

`syncState` now derives the quality label from:
- `effectiveQuality()` for streaming songs (reflects current network)
- Actual cached quality key for stream-cached songs (shows what's really cached)
- Original `qualityLabel` for downloaded songs

## Related

- #86 — Original bug report
- #96 — Previous fix attempt (shuffle index + ShuffleOrder resync) — merged but insufficient
- #95 — CoverArtEndpointInterceptor crash (fixed in #96)

## Tested

- Shuffle + WiFi → cellular → skip next: quality switches to 320kbps ✅
- Shuffle + cellular → WiFi → skip next: quality switches to Original ✅
- Sequential playback: same behavior ✅
- Stream cache: songs cache correctly, show correct cached quality ✅
- Cached songs on different network: show cached quality, not network quality ✅

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/playback/DefaultPlaybackManager.kt
+++ b/app/src/main/java/com/anzupop/saki/android/playback/DefaultPlaybackManager.kt
@@ -91,69 +91,6 @@ class DefaultPlaybackManager @Inject constructor(
         return if (cachedKey != null) StreamQuality.fromStorageKey(cachedKey) else preferred
     }
 
-    private suspend fun rebuildUpcomingItems() {
-        withController { activeController ->
-            val current = activeController.currentMediaItemIndex
-            if (current == C.INDEX_UNSET) return@withController
-            val count = activeController.mediaItemCount
-
-            val upcoming = shuffleDisplayOrder?.let { order ->
-                val pos = order.indexOf(current)
-                if (pos == -1) return@let null
-                val tail = order.drop(pos + 1)
-                // Wrap around for repeat-all so tracks after the end are also rebuilt
-                if (tail.size < 10 && activeController.repeatMode == Player.REPEAT_MODE_ALL) {
-                    (tail + order.take(10 - tail.size)).take(10)
-                } else {
-                    tail.take(10)
-                }
-            } ?: run {
-                val next = (current + 1).coerceAtMost(count)
-                val end = (current + 11).coerceAtMost(count)
-                val tail = (next until end).toList()
-                if (tail.size < 10 && activeController.repeatMode == Player.REPEAT_MODE_ALL) {
-                    (tail + (0 until (10 - tail.size).coerceAtMost(next))).take(10)
-                } else {
-                    tail
-                }
-            }
-
-            var replaced = false
-            for (i in upcoming) {
-                if (i !in 0 until count) continue
-                val item = activeController.getMediaItemAt(i)
-                val request = item.toPlaybackRequestOrNull() ?: continue
-                if (request.isCached) continue
-                val quality = resolveQualityForSong(request.serverId, request.songId)
-                if (quality.maxBitRate == request.maxBitRate) continue
-                val rebuilt = request.copy(
-                    qualityLabel = quality.label,
-                    streamCacheKey = streamCacheRepository.buildCacheKey(request.serverId, request.songId, quality),
-                    maxBitRate = quality.maxBitRate,
-                    format = quality.format,
-                ).let { updated ->
-                    MediaItem.Builder()
-                        .setMediaId(updated.songId)
-                        .setMediaMetadata(updated.toMediaMetadata())
-                        .setRequestMetadata(
-                            MediaItem.RequestMetadata.Builder()
-                                .setExtras(updated.toBundle())
-                                .build(),
-                        )
-                        .build()
-                }
-                activeController.replaceMediaItem(i, rebuilt)
-                replaced = true
-            }
-
-            // replaceMediaItem internally does remove+insert which corrupts ExoPlayer's
-            // ShuffleOrder. Re-send our deterministic order to restore it.
-            if (replaced && shuffleDisplayOrder != null) {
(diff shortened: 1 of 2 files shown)


xiongnemo (approved):
Image

PR #104, opened 2026-04-23
fix: seek bar no longer fights with playback progress during drag

### LINKED ISSUE #103: Now Playing: seek bar fights with playback progress during drag
## Bug

When dragging the seek bar in Now Playing, the thumb position fights with the player's real-time progress updates. The bar visibly jumps back and forth between the user's drag position and the current playback position, making precise seeking difficult.

## Expected

While the user is actively dragging (touch down → touch up), playback progress updates should be suppressed so the seek bar stays at the drag position.

## Likely cause

The seek bar value is bound to `positionMs` from `PlaybackSessionState`, which keeps updating via the position polling loop while the user is dragging. The drag state isn't being used to gate progress updates.

PengDingkang:
Closes #103

The seek bar slider value was bound to `positionMs` via a `LaunchedEffect` that fired continuously — even while the user was dragging. This caused the thumb to jump between the drag position and the real playback position.

Fix: add an `isDragging` flag that suppresses position updates during drag. Set on `onValueChange`, cleared on `onValueChangeFinished`.

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/presentation/library/PlayerChrome.kt
+++ b/app/src/main/java/com/anzupop/saki/android/presentation/library/PlayerChrome.kt
@@ -322,9 +322,12 @@ fun NowPlayingOverlay(
         var sliderValue by remember(track.songId) {
             mutableFloatStateOf(playbackState.positionMs.toFloat())
         }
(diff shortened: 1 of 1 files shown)


xiongnemo (approved):
Image

PR #105, opened 2026-04-23
fix: Now Playing album/artist tap navigates correctly from Settings

### LINKED ISSUE #99: Now Playing: tapping album navigates to Settings instead of album detail
## Bug

Tapping the album name in Now Playing navigates to Settings instead of the album detail page.

## Reproduction

1. Play a song
2. Navigate to Settings
3. From Settings, tap the mini player / Now Playing bar to open Now Playing
4. In Now Playing, tap the album name

**Expected**: Navigate to the album detail page
**Actual**: Navigates back to Settings

## Likely cause

The album tap action probably calls `navController.navigate(albumRoute)` which pushes onto the current back stack. Since the current back stack root is Settings, the navigation may be popping back to Settings instead of switching to the Browse tab where album detail lives.

PengDingkang:
Closes #99

## Problem

When opening Now Playing from Settings and tapping album/artist, the user was navigated back to Settings instead of the detail page. This happened because `showSettings` was never set to `false`, so `BrowseScreen` (which renders album/artist detail) was not displayed.

## Fix

In the `onOpenAlbum` and `onOpenArtist` callbacks of `NowPlayingOverlay`:
1. Set `showSettings = false` to switch underlying screen to Browse
2. Set the target album/artist state
3. Dismiss Now Playing overlay **last**

This ordering ensures the bottom screen is already showing the correct detail page when the overlay animates away — single smooth transition instead of two visible steps.

## Scope

Scanned all navigation callbacks from Now Playing and other overlays. Only album and artist taps are affected — queue sheet, lyrics, song details, and endpoint status have no outbound navigation.

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/presentation/SakiApp.kt
+++ b/app/src/main/java/com/anzupop/saki/android/presentation/SakiApp.kt
@@ -331,12 +331,14 @@ private fun RootShell(
             onDismiss = onDismissNowPlaying,
             canOpenArtist = canOpenArtistFromNowPlaying,
             onOpenArtist = {
-                onDismissNowPlaying()
+                showSettings = false
                 onOpenArtistFromPlayback(track.serverId, track.artistId)
+                onDismissNowPlaying()
             },
             onOpenAlbum = {
-                onDismissNowPlaying()
+                showSettings = false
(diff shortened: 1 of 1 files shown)


xiongnemo (approved):
Image

PR #110, opened 2026-04-23
fix: detect stream cache for transcoded streams without Content-Length

### LINKED ISSUE #109: Stream cache: transcoded streams never detected as cached (missing Content-Length)
## Bug

Transcoded streams (e.g. 320 kbps, 256 kbps) are never detected as stream-cached, even after the entire track has been buffered. The UI never shows "Cached" status for these tracks.

## Root cause

`DefaultStreamCacheRepository.buildSnapshot()` checks `isCached(key, 0, contentLength)` to determine if a track is fully cached. However, Subsonic's `stream` endpoint uses chunked transfer encoding for transcoded streams, which means **no Content-Length header** is sent. When `contentLength == C.LENGTH_UNSET`, the check is skipped entirely and the track is never marked as cached.

Original quality streams work because the server sends the file directly with a known Content-Length.

## Fix

When Content-Length is unknown, fall back to checking if the cache contains a single contiguous span starting at position 0 with non-zero size — this indicates the entire stream was cached from start to finish.

```kotlin
val isFullyCached = if (contentLength != C.LENGTH_UNSET.toLong() && contentLength > 0L) {
    streamCache.isCached(key, 0, contentLength)
} else {
    val spans = streamCache.getCachedSpans(key)
    spans.size == 1 && spans.first().position == 0L && cachedBytes > 0L
}
```

## Impact

- Stream-cached quality label in Now Playing never shows for transcoded tracks
- `findCachedQualityKey()` never finds transcoded cached tracks, so adaptive quality resolver doesn't know they're cached
- Settings page stream cache count undercounts

PengDingkang:
Closes #109

## Problem

Transcoded streams (320 kbps, 256 kbps, etc.) were never detected as stream-cached. The UI never showed "Cached" for these tracks, `findCachedQualityKey()` couldn't find them, and the stream cache count in Settings was undercounted.

## Root cause

`buildSnapshot()` required a known `Content-Length` to verify full caching via `isCached(key, 0, contentLength)`. Subsonic uses chunked transfer encoding for transcoded streams — no Content-Length header — so the check was always skipped.

## Fix

When Content-Length is unknown, fall back to checking if the cache has a single contiguous span starting at position 0 with non-zero size. This indicates the entire stream was written sequentially and completely.

One file changed, 9 insertions.

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/data/repository/DefaultStreamCacheRepository.kt
+++ b/app/src/main/java/com/anzupop/saki/android/data/repository/DefaultStreamCacheRepository.kt
@@ -146,7 +146,15 @@ class DefaultStreamCacheRepository @Inject constructor(
             bytesByServer[parsed.serverId] = (bytesByServer[parsed.serverId] ?: 0L) + cachedBytes
 
             val contentLength = ContentMetadata.getContentLength(streamCache.getContentMetadata(key))
-            if (contentLength != C.LENGTH_UNSET.toLong() && contentLength > 0L && streamCache.isCached(key, 0, contentLength)) {
+            val isFullyCached = if (contentLength != C.LENGTH_UNSET.toLong() && contentLength > 0L) {
+                streamCache.isCached(key, 0, contentLength)
+            } else {
+                // No Content-Length (e.g. transcoded stream): consider cached if there is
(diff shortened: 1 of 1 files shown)


xiongnemo (approved):
LGTM

PR #112, opened 2026-04-23
feat: configurable image cache size

### LINKED ISSUE #111: feat: configurable image cache size
## Problem

Image disk cache is hardcoded at 64 MB, which is easily exhausted with a moderately sized library (827 cached files at 66 MB on a test device with ~200 artists). Once full, Coil's LRU evicts older cover art, requiring re-download on revisit.

## Solution

Add a configurable image cache size setting, similar to the existing stream cache size setting.

- **Default**: 128 MB
- **Min**: 64 MB
- **Max**: 1024 MB (1 GB)
- **Step**: 64 MB

Image cache is pure disk I/O with no memory impact beyond Coil's separate memory cache (15% of app memory). Even at 1 GB, performance is unaffected — Coil manages LRU eviction internally and only reads from disk on memory cache miss.

## Implementation

- Add `imageCacheSizeMb` to `PlaybackPreferences` (or `AppPreferences`)
- DataStore key + settings UI slider (same pattern as stream cache size)
- Apply in `SakiApplication.newImageLoader()` — read preference synchronously at app start
- Changes require app restart (same as buffer strategy)

## Related

- Stream cache size setting (existing pattern to follow)

PengDingkang:
Closes #111

## Problem

Image disk cache was hardcoded at 64 MB, easily exhausted with a moderately sized library (~800 cached files). Once full, Coil evicts older cover art requiring re-download on revisit.

## Solution

Add a configurable cover art cache size slider in Settings → Offline audio:

- **Default**: 128 MB (doubled from previous 64 MB)
- **Range**: 64 MB – 1 GB
- **Step**: 64 MB
- Restart required to apply (snackbar notification)

### Files changed

- `PlaybackModels.kt` — constants + `imageCacheSizeMb` field
- `PlaybackPreferencesRepository.kt` — `updateImageCacheSizeMb`
- `DataStorePlaybackPreferencesRepository.kt` — key, normalize, mapping
- `DefaultPlaybackPreferencesRepository.kt` — error stub
- `SakiApplication.kt` — read preference for Coil `DiskCache.maxSizeBytes`
- `SakiAppViewModel.kt` — update method with snackbar
- `SettingsScreen.kt` — slider UI
- `SakiApp.kt` — callback wiring


xiongnemo (approved):
Image

PR #118, opened 2026-04-24
fix: notification artwork resilient to network/endpoint changes

### LINKED ISSUE #117: Notification cover art doesn't update after network switch
## Bug

After switching networks or endpoints, the cover art in the media notification / status bar / lock screen fails to load. In-app cover art works fine.

## Root cause

`MediaMetadata.artworkUri` is set to a **concrete HTTP URL** tied to a specific endpoint (e.g. `http://192.168.1.100:44533/rest/getCoverArt?id=xxx`) at queue build time in `resolvePlayableItem`. The system's media notification renderer loads this URL directly using its own HTTP client — **not through our OkHttp/CoverArtEndpointInterceptor pipeline**.

When the network changes (WiFi→cellular), the LAN endpoint becomes unreachable, and the system silently fails to load the artwork.

In-app artwork works because Coil goes through `CoverArtEndpointInterceptor`, which rewrites URLs to the current best endpoint.

### 1. Embed bitmap via `artworkData` in `onAddMediaItems`
Load artwork bitmap via Coil in `resolvePlayableItem` and set `MediaMetadata.artworkData` instead of `artworkUri`.

**Result**: Broke playback. `onAddMediaItems` must return quickly — Coil's network fetch blocks the queue addition, causing Media3 to timeout.

### 2. Async bitmap via `NotificationArtworkLoader` listener
Load bitmap on `onMediaItemTransition`, then `replaceMediaItem` with updated metadata containing `artworkData`.

**Result**: 
- `replaceMediaItem` triggers `onMediaItemTransition` → infinite loop (mitigated with guard)
- `replaceMediaItem` corrupts `ShuffleOrder` in shuffle mode (same root cause as #86)
- Even with `artworkUri(null)` + `artworkData(bytes)`, system notification still appeared to use the original URL

### 3. Set `artworkUri(null)` + `artworkData` via `replaceMediaItem`
Clear the HTTP URL and only provide embedded bytes.

**Result**: System notification still showed delayed loading behavior, suggesting Media3 or the system ignores `artworkData` in favor of its own artwork loading mechanism.

## Constraints

| Requirement | Why |
|---|---|
| `onAddMediaItems` must be fast | Media3 times out if it blocks |
| Can't use `replaceMediaItem` | Corrupts ShuffleOrder (#86) |
| System notification loads `artworkUri` directly | Doesn't use our OkHttp/Coil |
| `artworkData` may be ignored | System/Media3 may prefer URI |

## Possible solutions to explore

1. **Custom `MediaNotification.Provider`**: Take full control of notification building, load artwork via Coil ourselves, set bitmap directly on the notification
2. **`MediaSessionService.setMediaNotificationProvider`**: Override how Media3 builds the notification
3. **Local content:// URI**: Instead of HTTP URL, write artwork to a local file and use a `content://` URI via FileProvider — system can always load local files
4. **Preload artwork to disk in background**: When queue is built, kick off async Coil preloads. By the time the track plays, artwork is in disk cache. Use a `content://` URI pointing to the cache file.

## Related

- #95 — CoverArtEndpointInterceptor crash (fixed)
- #86 — replaceMediaItem corrupts ShuffleOrder
- #45 — Multi-endpoint smart selection

PengDingkang:
Closes #117

## Problem

Notification/lock screen cover art failed to load after network or endpoint changes. In-app artwork worked fine.

The system's media notification renderer loaded `artworkUri` (a baked HTTP URL tied to a specific endpoint) directly, bypassing our OkHttp pipeline and `CoverArtEndpointInterceptor`. After network switch, the URL became unreachable.

### 1. `CoilBitmapLoader` (new)
Custom `BitmapLoader` set on `MediaSession.Builder.setBitmapLoader`. Media3 now loads notification artwork through Coil → OkHttp → `CoverArtEndpointInterceptor`, which rewrites the URL to the current best endpoint.

- In-memory LRU cache (20 bitmaps) for instant repeat loads
- Retry with 2s delay on failure to allow `EndpointSelector` reprobe after network change

### 2. Canonical endpoint for `artworkUri`
`resolvePlayableItem` now builds `artworkUri` using the canonical endpoint (first by order) instead of the current best endpoint. This ensures `CoverArtEndpointInterceptor` recognizes and rewrites the URL regardless of which endpoint was active at queue build time.

### Data flow
```
artworkUri (canonical endpoint, stable)
  → CoilBitmapLoader.loadBitmap()
    → Coil (memory cache → disk cache → network)
      → OkHttp + CoverArtEndpointInterceptor (rewrites to current best endpoint)
        → notification bitmap
```

## Failed approaches (documented in #117)
- Embedding bitmap in `onAddMediaItems` → blocked queue addition
- `replaceMediaItem` with `artworkData` → corrupted ShuffleOrder
- `artworkUri(null)` + `artworkData` → system ignored embedded data

### DIFF
--- a/app/src/main/java/com/anzupop/saki/android/playback/CoilBitmapLoader.kt
+++ b/app/src/main/java/com/anzupop/saki/android/playback/CoilBitmapLoader.kt
@@ -1,0 +1,80 @@
+package com.anzupop.saki.android.playback
+
+import android.content.Context
+import android.graphics.Bitmap
+import android.graphics.BitmapFactory
+import android.net.Uri
+import androidx.media3.common.util.UnstableApi
+import androidx.media3.session.BitmapLoader
+import coil3.SingletonImageLoader
+import coil3.request.ImageRequest
+import coil3.request.SuccessResult
+import coil3.toBitmap
+import com.google.common.util.concurrent.Futures
+import com.google.common.util.concurrent.ListenableFuture
+import com.google.common.util.concurrent.SettableFuture
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.launch
+
+/**
+ * A [BitmapLoader] that uses Coil to load artwork bitmaps. This ensures notification
+ * artwork goes through our OkHttp pipeline (including [CoverArtEndpointInterceptor]),
+ * so artwork loads correctly after network/endpoint changes.
+ */
+@UnstableApi
+class CoilBitmapLoader(private val context: Context) : BitmapLoader {
+
+    private val scope = CoroutineScope(SupervisorJob() + Dispatchers.Main)
+    private val bitmapCache = android.util.LruCache<String, Bitmap>(20)
+
+    override fun supportsMimeType(mimeType: String): Boolean = mimeType.startsWith("image/")
+
+    override fun decodeBitmap(data: ByteArray): ListenableFuture<Bitmap> {
+        val future = SettableFuture.create<Bitmap>()
+        try {
+            val bitmap = BitmapFactory.decodeByteArray(data, 0, data.size)
+            if (bitmap != null) future.set(bitmap) else future.setException(IllegalArgumentException("Failed to decode bitmap"))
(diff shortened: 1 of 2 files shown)


xiongnemo (approved):
image

PR #196, opened 2026-04-27
fix: defer large playback queue loading

### LINKED ISSUE #190: Large queue playback delay when playing from Songs tab
## Problem
Playing a song from the Songs tab (which contains the full library, potentially thousands of songs) has a noticeable delay before playback starts. Playing from an album (10-20 songs) is instant.

## Root cause
`DefaultPlaybackManager.playQueue()` blocks on two expensive operations before playback begins:

1. **`songs.map { toPreferredMediaItem() }`** — calls `findCachedQualityKey()` per song (Set lookups × quality entries × thousands of songs)
2. **`controller.setMediaItems(mediaItems)`** — Media3 processes thousands of MediaItems internally

Both complete before `play()` is called.

## Proposed fix: instant-play with deferred queue

1. Immediately build a single MediaItem for the tapped song → `setMediaItem` → `prepare` → `play` (instant response)
2. In a background coroutine, build the full `mediaItems` list
3. Once ready, call `setMediaItems` with the complete queue, preserving the current playback position
4. If shuffle is enabled, apply `SakiShuffleOrder` after the full queue is set

This keeps shuffle working correctly on the full queue while eliminating perceived latency. `SakiShuffleOrder` only depends on queue length + seed, so it's safe to apply after the deferred queue replacement.

## Scope
- `DefaultPlaybackManager.playQueue()`
- Affects Songs tab and any future large-list playback (search results, cached songs list)

MYmeru:
## Summary
- Start large `playQueue` requests by immediately playing the selected song
- Build the full queue in the background and insert items around the current track
- Preserve/cancel deferred queue loads safely across later queue mutations
- Keep shuffle behavior correct by applying `SakiShuffleOrder` after the full queue is ready, including pending shuffle toggles during queue expansion

Closes #190

### DIFF
--- a/app/src/main/java/org/hdhmc/saki/playback/DefaultPlaybackManager.kt
+++ b/app/src/main/java/org/hdhmc/saki/playback/DefaultPlaybackManager.kt
@@ -74,6 +74,8 @@ class DefaultPlaybackManager @Inject constructor(
     private var deferredQueueJob: Job? = null
     private var queueLoadGeneration: Long = 0L
     private var pendingDeferredShuffleEnabled: Boolean? = null
+    private var pendingDeferredShuffleSeed: Long = 0L
+    private var pendingDeferredShuffleAnchorIndex: Int = 0
 
     private val controllerListener = object : Player.Listener {
         override fun onEvents(player: Player, events: Player.Events) {
@@ -208,9 +210,15 @@ class DefaultPlaybackManager @Inject constructor(
             val shouldRestoreShuffle = shuffleDisplayOrder != null || activeController.shuffleModeEnabled
             val deferredShuffleSeed = if (shouldRestoreShuffle) System.nanoTime() else 0L
             pendingDeferredShuffleEnabled = if (songs.size > 1) shouldRestoreShuffle else null
+            pendingDeferredShuffleSeed = deferredShuffleSeed
+            pendingDeferredShuffleAnchorIndex = safeStartIndex
             shuffleDisplayOrder = null
             activeController.shuffleModeEnabled = false
-            if (!shouldRestoreShuffle || songs.size == 1) {
+            if (shouldRestoreShuffle && songs.size > 1) {
+                shuffleSeed = deferredShuffleSeed
+                shuffleAnchorIndex = safeStartIndex
+                persistShuffleState(shuffleSeed, shuffleAnchorIndex)
+            } else {
                 persistShuffleState()
             }
             activeController.setMediaItem(startMediaItem)
@@ -235,6 +243,8 @@ class DefaultPlaybackManager @Inject constructor(
         deferredQueueJob?.cancel()
         deferredQueueJob = null
         pendingDeferredShuffleEnabled = null
+        pendingDeferredShuffleSeed = 0L
+        pendingDeferredShuffleAnchorIndex = 0
         queueLoadGeneration += 1
         return queueLoadGeneration
     }
@@ -285,9 +295,14 @@ class DefaultPlaybackManager @Inject constructor(
                     val fullQueueSize = activeController.mediaItemCount
                     val currentIndex = activeController.currentMediaItemIndex
                     val enableShuffle = pendingDeferredShuffleEnabled ?: restoreShuffle
+                    val pendingShuffleSeed = pendingDeferredShuffleSeed
                     pendingDeferredShuffleEnabled = null
+                    pendingDeferredShuffleSeed = 0L
+                    pendingDeferredShuffleAnchorIndex = 0
                     if (enableShuffle && fullQueueSize > 1 && currentIndex in 0 until fullQueueSize) {
-                        shuffleSeed = shuffleSeedForQueue.takeUnless { it == 0L } ?: System.nanoTime()
+                        shuffleSeed = pendingShuffleSeed.takeUnless { it == 0L }
+                            ?: shuffleSeedForQueue.takeUnless { it == 0L }
+                            ?: System.nanoTime()
                         shuffleAnchorIndex = currentIndex
                         sendShuffleOrderToService(activeController, fullQueueSize, shuffleSeed, shuffleAnchorIndex)
                         activeController.shuffleModeEnabled = true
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/hdhmc/saki/playback/DefaultPlaybackManager.kt
```diff
@@ -185,35 +195,188 @@ class DefaultPlaybackManager @Inject constructor(
...
+                cachedSong = cachedSongsById[song.id],
+            )
         }
+        DeferredQueueItems(before = before, after = after)
     }
 
+    private data class DeferredQueueItems(
+        val before: List<MediaItem>,
+        val after: List<MediaItem>,
+    )
```
PengDingkang, line 372: This `finally` block is effectively dead code. When the deferred job is cancelled, `beginDeferredQueueLoad()` has already incremented `queueLoadGeneration`, so `generation != queueLoadGeneration` and the block is skipped. When the job completes normally, the `withController` block above already calls `persistShuffleState()` and clears the pending state.

Consider removing this or adding a comment explaining what scenario it's meant to cover.

app/src/main/java/org/hdhmc/saki/playback/DefaultPlaybackManager.kt
```diff
@@ -66,6 +71,11 @@ class DefaultPlaybackManager @Inject constructor(
     private var shuffleAnchorIndex: Int = 0
     private var shuffleDisplayOrder: List<Int>? = null // player indices in shuffle display order
     private var restoringExternalShuffleState = false
+    private var deferredQueueJob: Job? = null
+    private var queueLoadGeneration: Long = 0L
```
PengDingkang, line 75: 5 new mutable state variables (`deferredQueueJob`, `queueLoadGeneration`, `pendingDeferredShuffleEnabled`, `pendingDeferredShuffleSeed`, `pendingDeferredShuffleAnchorIndex`) are read/written across `playQueue`, `scheduleDeferredQueueLoad`, `toggleShuffle`, `cancelDeferredQueueLoad`, and `syncState`. While thread-safe (all on main dispatcher), the state transitions are hard to follow.

Non-blocking suggestion: consider encapsulating these into a `DeferredQueueState` data class with explicit transition methods, making the valid state combinations self-documenting. Can be a follow-up.

app/src/main/java/org/hdhmc/saki/playback/DefaultPlaybackManager.kt
```diff
@@ -185,35 +195,188 @@ class DefaultPlaybackManager @Inject constructor(
         startIndex: Int,
     ) {
         require(songs.isNotEmpty()) { "Playback queue cannot be empty." }
-        shuffleDisplayOrder = null
-        persistShuffleState()
+        val safeStartIndex = startIndex.coerceIn(songs.indices)
+        val startSong = songs[safeStartIndex]
         val preferredQuality = effectiveQuality()
+        val startCachedSong = cachedSongRepository.getPlayableCachedSong(serverId, startSong.id, preferredQuality)
```
PengDingkang, line 201: Is `getPlayableCachedSong` (singular) a new method? The existing codebase only has `getPlayableCachedSongs` (batch). If this was added as part of this PR, it's not in the diff — please confirm it exists in `CachedSongRepository`.

app/src/main/java/org/hdhmc/saki/playback/DefaultPlaybackManager.kt
```diff
@@ -408,6 +580,23 @@ class DefaultPlaybackManager @Inject constructor(
...
+                        val seed = pendingDeferredShuffleSeed.takeUnless { it == 0L } ?: System.nanoTime()
+                        pendingDeferredShuffleSeed = seed
+                        shuffleSeed = seed
+                        shuffleAnchorIndex = pendingDeferredShuffleAnchorIndex
+                        persistShuffleState(seed, pendingDeferredShuffleAnchorIndex)
+                    } else {
+                        persistShuffleState()
+                    }
+                    syncState(activeController)
+                    return@withController
```
PengDingkang, line 602: The `count <= 1` guard inside the `pendingDeferredShuffleEnabled?.let` block relies on the fact that if the deferred job has already inserted items (`count > 1`), we fall through to the normal shuffle path. This works but is fragile — if the deferred job completes `addMediaItems` but hasn't yet cleared `pendingDeferredShuffleEnabled`, we enter the `let` block, skip the `count <= 1` branch, and fall through. Correct behavior by coincidence.

A comment here explaining the intended flow would help future readers.

PengDingkang (commented):
Core approach is solid — instant playback of the tapped song with deferred queue expansion solves the latency issue correctly. Shuffle handling during the deferred window is thorough.

Main concerns:
1. **`finally` block appears to be dead code** — generation check prevents execution in all reachable paths
2. **Shuffle pending state complexity** — 5 interrelated mutable vars across 5 methods; consider encapsulating (follow-up OK)
3. **`getPlayableCachedSong` (singular)** — not in the diff, please confirm it exists
4. **`toggleShuffle` pending path** — correct by coincidence, needs a comment

None of these block merge after the `getPlayableCachedSong` confirmation. Nice work on the cancellation safety and generation tracking.

app/src/main/java/org/hdhmc/saki/playback/DefaultPlaybackManager.kt
MYmeru: I added comments for the deferred cleanup path and pending-shuffle toggle path. I’d keep the `DeferredQueueState` encapsulation as a follow-up to avoid expanding the behavioral surface in this fix.

app/src/main/java/org/hdhmc/saki/playback/DefaultPlaybackManager.kt
MYmeru: Confirmed `getPlayableCachedSong` already exists in `CachedSongRepository` and `DefaultCachedSongRepository`; it is not introduced by this PR.

PengDingkang (approved):

PR #243, opened 2026-05-10
feat: timeline-aware custom buffer prefetch

### LINKED ISSUE #242: Implement timeline-aware custom buffer prefetch
## Context

#235 / #238 moved CUSTOM long buffering away from ExoPlayer's Java heap by keeping the player buffer short and prefetching the current stream into SimpleCache.

That fixes the OOM class, but the user-facing setting still needs clearer end-to-end semantics: the CUSTOM buffer duration should mean "buffer this much playable timeline ahead", not "buffer only this much of the current track". If the remaining current track is shorter than the configured duration, prefetch should continue into the next track(s) in queue order.

## Target behavior

For CUSTOM buffer duration D:

- Keep ExoPlayer's in-memory player buffer short and heap-bounded as fixed by #238.
- Use disk prefetch to fill up to D milliseconds of continuous playback timeline ahead of the current position.
- First fill the current track from the current position to the end, then continue with the next playable queue item, then later items until the configured duration budget is covered or the queue cannot continue.
- Respect repeat and shuffle semantics exactly as the player queue exposes them.
- Repeat-one should not aggressively cache the same track forever; it can stop after the current track target is satisfied.
- If offline-degraded, unavailable, local cached, or already fully stream-cached, skip or stop according to the existing playback rules.

## Implementation sketch

- Introduce a small prefetch planner that reads the current player queue snapshot and produces ordered prefetch targets with remaining-duration budgets.
- Run one CacheWriter target at a time, with current-track work always taking priority over future-track work.
- Cancel and recompute on media item transition, queue mutation, shuffle/repeat changes, quality/network/server changes, and offline-degraded state changes.
- Apply the existing SimpleCache evictor and stream cache size limit; do not bypass user cache limits.
- Keep current-track buffered UI based on current track only. Future-track prefetch should not inflate the current progress bar.

## Acceptance criteria

- CUSTOM buffer duration includes upcoming tracks when the current track has less remaining time than the configured target.
- CUSTOM buffering does not regress the OOM fix from #238.
- Current-track prefetch remains prioritized over future-track work.
- Future-track prefetch is canceled when the queue target changes.
- NORMAL buffer behavior is unchanged.
- Shuffle, repeat-one, repeat-all, manual queue edits, virtual queue playback, server changes, and network-quality changes remain correct.

## Notes

This should stay separate from the #235 OOM fix because it touches queue semantics and has a wider regression surface.

PengDingkang:
Closes #242

## Summary

Extends CUSTOM buffer prefetch to cover the configured duration across multiple queue items, not just the current track.

## Changes

`SakiPlaybackService.kt`:

- **`buildStreamPrefetchPlan`** — walks the player timeline from `currentIndex` forward using `Timeline.getNextWindowIndex(index, repeatMode, shuffleModeEnabled)`, accumulating tracks until the configured buffer duration budget is exhausted. Produces an ordered list of `StreamPrefetchTarget`s.
- **`prefetchTimelineToDisk`** — iterates targets serially, calling `prefetchStreamToDisk` (single `CacheWriter`) for each uncached track.
- **`StreamPrefetchListener`** — now also reacts to `onTimelineChanged`, `onRepeatModeChanged`, `onShuffleModeEnabledChanged` to recompute the plan.
- After a plan completes successfully, re-evaluates to pick up newly-eligible tracks (e.g. playback advanced).

## Behavior

| Scenario | Behavior |
|----------|----------|
| Current track shorter than configured duration | Prefetch continues into next queue items |
| Shuffle enabled | Uses player's shuffle order via `getNextWindowIndex` |
| Repeat-all | Wraps around queue; stops when budget exhausted |
| Repeat-one | Only prefetches current track, then stops |
| Track already cached / local / offline-degraded | Skipped but duration still deducted from budget |
| Queue mutation / mode change | Plan key invalidates → cancel + recompute |

## Not changed

- Player in-memory buffer remains short and heap-bounded (#238)
- NORMAL strategy unchanged
- Progress bar still reflects current-track cache only (per #242 spec)
- Single `CacheWriter` at a time, no concurrent disk/network load

### DIFF
--- a/app/src/main/java/org/hdhmc/saki/data/repository/DefaultStreamCacheRepository.kt
+++ b/app/src/main/java/org/hdhmc/saki/data/repository/DefaultStreamCacheRepository.kt
@@ -169,10 +169,10 @@ class DefaultStreamCacheRepository @Inject constructor(
             val isFullyCached = if (contentLength != C.LENGTH_UNSET.toLong() && contentLength > 0L) {
                 streamCache.isCached(key, 0, contentLength)
             } else {
-                // No Content-Length (e.g. transcoded stream): consider cached if there is
-                // a single contiguous span starting at 0 with meaningful size
-                val spans = streamCache.getCachedSpans(key)
-                spans.size == 1 && spans.first().position == 0L && cachedBytes > 0L
+                // No Content-Length (e.g. transcoded stream): consider cached if the
+                // resource has a contiguous cached prefix starting at 0. SimpleCache
+                // may split a completed CacheWriter result into multiple spans.
+                hasContiguousCachedPrefix(key)
             }
             if (isFullyCached) {
                 cachedSongIdsByServerAndQuality
@@ -188,6 +188,22 @@ class DefaultStreamCacheRepository @Inject constructor(
             },
             bytesByServer = bytesByServer.toMap(),
         )
+    }
+
+    private fun hasContiguousCachedPrefix(key: String): Boolean {
+        val spans = streamCache.getCachedSpans(key).sortedBy { span -> span.position }
+        if (spans.isEmpty() || spans.first().position != 0L) return false
+
+        var coveredUntil = 0L
+        for (span in spans) {
+            if (span.length <= 0L) continue
+            if (span.position > coveredUntil) return false
+            val spanEnd = span.position + span.length
+            if (spanEnd > coveredUntil) {
+                coveredUntil = spanEnd
+            }
+        }
+        return coveredUntil > 0L
     }
 }
 
--- a/app/src/main/java/org/hdhmc/saki/playback/SakiPlaybackService.kt
+++ b/app/src/main/java/org/hdhmc/saki/playback/SakiPlaybackService.kt
@@ -136,7 +136,7 @@ class SakiPlaybackService : MediaSessionService() {
     private var streamPrefetchJob: Job? = null
     private var streamPrefetchReevaluationJob: Job? = null
     private var activeStreamPrefetchKey: String? = null
-    private val completedStreamPrefetchKeys = ConcurrentHashMap.newKeySet<String>()
+    private val completedStreamPrefetchKeysByTarget = ConcurrentHashMap<StreamPrefetchTargetKey, String>()
     @Volatile
     private var activeStreamCacheWriter: CacheWriter? = null
 
@@ -560,13 +560,20 @@ class SakiPlaybackService : MediaSessionService() {
             if (!request.isCached && request.localPath == null && !endpointSelector.isOfflineDegraded(request.serverId)) {
                 val quality = request.requestedStreamQuality(prefs)
                 val prefetchRequest = request.withStreamQuality(quality)
+                val targetKey = StreamPrefetchTargetKey(
+                    serverId = request.serverId,
+                    songId = request.songId,
+                    qualityKey = quality.storageKey,
+                )
+                val isSatisfied = isStreamPrefetchSatisfied(request.serverId, request.songId, quality)
                 keyParts +=
-                    "$index:${request.serverId}:${request.songId}:${quality.storageKey}:${quality.format.orEmpty()}"
-                if (!isStreamPrefetchSatisfied(request.serverId, request.songId, quality)) {
+                    "$index:${targetKey.planKey()}:${if (isSatisfied) "done" else "pending"}"
+                if (!isSatisfied) {
                     targets += StreamPrefetchTarget(
                         request = prefetchRequest,
                         quality = quality,
                         queueIndex = index,
+                        targetKey = targetKey,
                     )
                 }
             }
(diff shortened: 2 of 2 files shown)


PengDingkang:
## Bug: prefetch completes but cache lookup misses

### Symptom

After the initial prefetch plan completes (track N and N+1 downloaded), subsequent reevaluations still report `alreadyCached=false` for both tracks. The prefetch job restarts and finishes in ~6ms (data already in SimpleCache), but `findCachedQualityKey` never finds it. This means:

1. Track N+2 never enters the plan (budget exhausted by N+1 which is "not cached")
2. On natural track transition, the next track shows only ~60s of player buffer instead of full disk cache

### Debug log evidence

```
01:41:02.091 prefetchStreamToDisk: song=h2zQoSq2XhLceNTxVSMw6u
01:41:02.136 prefetchStreamToDisk: song=5VWqMVnn3BFaKrFjfl1qOC
01:41:02.856 prefetch job finished, success=true

// 4 seconds later — both still "not cached":
01:41:06.778 [40] alreadyCached=false quality=original
01:41:06.779 [41] alreadyCached=false quality=original
01:41:06.779 buildPlan result: targets=0 keyParts=2 budgetLeft=-137601ms

// Next attempt finishes in 6ms (data already on disk):
01:41:03.057 prefetchStreamToDisk: song=h2zQoSq2XhLceNTxVSMw6u
01:41:03.063 prefetchStreamToDisk: song=5VWqMVnn3BFaKrFjfl1qOC
01:41:03.071 prefetch job finished, success=true
```

### Root cause

**Cache key mismatch** between:
- `buildStreamPrefetchPlan` → calls `findCachedQualityKey(serverId, songId, quality)` which builds a lookup key via `streamCacheRepository.buildCacheKey(serverId, songId, quality)`
- `prefetchStreamToDisk` → calls `resolveStreamDataSpec` which resolves the actual stream URL and sets `dataSpec.key` — this is the key `CacheWriter` writes to `SimpleCache`

These two keys are constructed differently. The prefetch writes data under the resolved key, but the plan checks a different key format, so it never finds the cached data.

### Secondary issue

Even if the cache key is fixed, there's a missing periodic reevaluation: once all targets in a plan are satisfied, the plan won't expand to cover further tracks (e.g. track N+2) until a discrete event (track transition, mode change) triggers `syncCurrentStreamPrefetch`. A 30s reevaluation timer is needed but is secondary to the key mismatch fix.

### Fix direction

Align the cache key used in `buildStreamPrefetchPlan`'s `findCachedQualityKey` check with the key that `resolveStreamDataSpec` produces for `CacheWriter`. Either:
1. Use `streamCacheRepository.buildCacheKey` in `resolveStreamDataSpec` (if not already), or
2. Use the resolved spec key in the plan's cache check

PengDingkang:
## Bug confirmed: infinite prefetch loop after `5d71d65`

### Evidence

Song `EfKp8nlWXizbigRP83C5mD` (index 44) is prefetched in a tight loop (~250ms per iteration, ~4 downloads/second):

```
10:40:14.839 [44] prefetching song=EfKp8nlWXizbigRP83C5mD
10:40:15.212 [44] completed, cacheKey=saki.stream.v1|1|EfKp8nlWXizbigRP83C5mD|
10:40:15.214 sync: starting prefetch, targets=1
10:40:15.214 [44] prefetching song=EfKp8nlWXizbigRP83C5mD
10:40:15.502 [44] completed, cacheKey=saki.stream.v1|1|EfKp8nlWXizbigRP83C5mD|
10:40:15.503 sync: starting prefetch, targets=1
... (repeats indefinitely)
```

This is the source of the 60MB/s network traffic.

### Root cause

Two issues combine:

1. **`completedStreamPrefetchKeys` check fails for index 44.** The plan's `keyParts` only includes indices 42 and 43 (the tracks that were in the plan when it was first built). Index 44 enters the plan via reevaluation as budget frees up, but `isStreamPrefetchSatisfied` doesn't recognize it as completed because `findCompletedStreamPrefetchKey` uses `streamCacheRepository.buildCacheKey(serverId, songId, quality)` which produces a different key format than what `CacheWriter` actually wrote (`saki.stream.v1|1|...|`).

2. **Plan key doesn't change between iterations.** The key is `360000:0:false:42:...|43:...` — it doesn't include index 44. So `activeStreamPrefetchKey == plan.key` is true, but `streamPrefetchJob?.isActive` is false (job just finished). This means the "same key, skip" guard doesn't fire, and a new job starts immediately.

### The `completedStreamPrefetchKeys` workaround doesn't work because:

`findCompletedStreamPrefetchKey` calls `streamCacheRepository.buildCacheKey(serverId, songId, quality)` which returns something like `saki.stream.v1|1|EfKp8nlWXizbigRP83C5mD|original` (with quality suffix), but the actual cache key from `resolveStreamDataSpec` is `saki.stream.v1|1|EfKp8nlWXizbigRP83C5mD|` (without quality, or with different format). The exact key comparison fails.

### Fix needed

The `completedStreamPrefetchKeys` set stores the **actual** resolved cache key (from `resolvedSpec.key`). The lookup in `isCompletedStreamPrefetchKey` should compare against the same resolved key, not a reconstructed one. Simplest fix: store `(serverId, songId, quality)` tuples instead of raw cache keys, or just use the song ID as the dedup key since we're already filtering by quality in the plan.

PengDingkang:
## Bug #3: CacheWriter writes 0 bytes for certain tracks

### Evidence

```
11:26:21.243 prefetch start: key=saki.stream.v1|1|T4qkKEmb0xfwoX0xJvZ9u1|original
11:26:21.243 existing cache: spans=0 bytes=0
11:26:21.590 prefetch done: spans=1 bytes=10955030  ← SUCCESS (10 MB)

11:26:52.432 prefetch start: key=saki.stream.v1|1|KfrTuz8M5F3UG8bU16wLXV|original
11:26:52.432 existing cache: spans=0 bytes=0
11:26:52.691 prefetch done: spans=0 bytes=0  ← FAILURE (0 bytes written!)
// repeats infinitely, ~250ms per attempt, each with different auth token
```

### Root cause

The Subsonic server returns **no Content-Length header** for this track (likely a transcoded stream using chunked transfer encoding). When `CacheDataSource` opens a request without a known content length, `SimpleCache` cannot allocate a cache span for writing. The `CacheWriter.cache()` call reads the response body (consuming bandwidth) but the data passes through without being written to disk.

This explains the 60MB/s traffic: the server streams the full track data each time, `CacheWriter` reads it all, but nothing lands in the cache. The `completedStreamPrefetchKeys` fix correctly records it as "done", but `isCompletedStreamPrefetchCacheKey` then checks `getCachedSpans` and finds 0 spans → removes from map → next iteration tries again.

### Why some tracks work and others don't

- Tracks served as raw files (no transcoding): server sends Content-Length → CacheWriter writes successfully
- Tracks requiring transcoding: server uses chunked encoding (no Content-Length) → CacheWriter reads but can't write

### Fix options

1. **Set `DataSpec.FLAG_ALLOW_CACHE_FRAGMENTATION`** on the prefetch DataSpec — allows SimpleCache to write without knowing total length upfront (writes in fragments)
2. **Skip prefetch for tracks where CacheWriter fails to write** — add the target to a "skip set" after detecting 0 bytes written, don't retry
3. **Both** — try with fragmentation flag, and if still 0 bytes, skip permanently for this session

PengDingkang:
## Root cause confirmed: SimpleCache write lock conflict with ExoPlayer

### Evidence

The last line of the log session shows the real pattern:
```
// Dozens of attempts: 0 bytes (ExoPlayer holds the lock)
11:35:12.898 prefetch start: key=saki.stream.v1|1|QUotwadhpYncut9abMcAvV|original
11:35:12.898 existing cache: spans=0 bytes=0
11:35:13.361 prefetch done: spans=1 bytes=5242880  ← finally wrote 5 MB (one fragment)
```

5242880 = 5 MB = `CacheDataSink.DEFAULT_FRAGMENT_SIZE`. The write only succeeds when ExoPlayer releases its lock on that cache key's span.

### Why "first 2 tracks OK, 3rd onwards fails"

ExoPlayer's player buffer is set to `CUSTOM_PLAYER_MAX_BUFFER_MS = 5 min`. When playback starts at track N:
- Track N: ExoPlayer is actively reading/writing this cache key → prefetch skips it (already `isCurrent`)
- Track N+1: ExoPlayer hasn't started loading it yet → prefetch succeeds immediately
- Track N+2: By the time prefetch reaches it, ExoPlayer's 5-min buffer has started loading it → **lock conflict** → CacheWriter reads the full response from server but SimpleCache rejects the write → 0 bytes → infinite retry

### Fix

**Skip tracks that ExoPlayer is currently buffering.** The prefetch planner should check if a track's cache key already has an active `CacheSpan` lock (via `SimpleCache.startReadWriteNonBlocking` or similar), or more simply: skip the first N tracks that fall within `CUSTOM_PLAYER_MAX_BUFFER_MS` of the current position, since ExoPlayer will handle those itself.

Simplest approach: in `buildStreamPrefetchPlan`, don't add targets for tracks whose cumulative offset from current position is ≤ `CUSTOM_PLAYER_MAX_BUFFER_MS`. Only prefetch tracks *beyond* what the player buffer covers.

MYmeru (approved):

PR #293, opened 2026-05-31
Shorten prefetch retry after a no-op write so CUSTOM buffer fills ahead

### LINKED ISSUE #292: CUSTOM buffer duration not achieved while streaming (current track never disk-prefetched / player byte cap)
## Bug

With CUSTOM buffer strategy, the configured buffer duration is not actually achieved during streaming — the effective buffered-ahead amount stays small, and playback hits the network again well before the configured duration. Reported as "custom buffer duration stops working after playing for a while", but it reproduces **before the stream cache is anywhere near full**, so LRU eviction (a 2 GB cache filling up) is **ruled out**.

## Likely root cause (to confirm with logging)

The disk-prefetch planner (`SakiPlaybackService.buildStreamPrefetchPlan`) marks any track whose end is within `CUSTOM_PLAYER_MAX_BUFFER_MS` (5 min) of the current position as `"player"` (assumed covered by the in-memory player buffer) and **does not disk-prefetch it**. Songs are almost always < 5 min, so the **current track is never disk-prefetched** — it relies entirely on ExoPlayer's in-memory buffer.

But that in-memory buffer is bounded by `SakiLoadControl`'s byte cap (`customTargetBufferBytes()` = heap/4, clamped 16–96 MB, with `prioritizeTimeOverSizeThresholds = false`). For lossless/high-bitrate audio (e.g. 1028 kbps FLAC, 16 MB ≈ ~2 min) the player buffer caps out far below the configured duration. Net effect: the configured custom buffer never materializes for the current track.

This contradicts #242's acceptance criterion: *"First fill the current track from the current position to the end, then continue with the next playable queue item…"*.

Also note: for `customBufferSeconds <= 300` (incl. the default 60 s), `buildStreamPrefetchPlan` returns early — there is **no disk prefetch at all**, so the whole buffer is the byte-capped player buffer.

## Plan

1. Add temporary, low-noise diagnostic logging (tag `SakiStreamPrefetch`) capturing per-`syncCurrentStreamPrefetch`: strategy, `customBufferSeconds`, current index/position/duration, player `bufferedPosition` and ahead-ms, `customTargetBufferBytes()`, `streamCache.cacheSpace`, and the plan key (per-track target states).
2. Reproduce, capture logcat, confirm A (small ahead-ms regardless of configured duration; current track not `pending`) vs any other cause.
3. Implement the targeted fix and drop/gate the logging.

## Related
- #235 (CUSTOM buffer OOM — introduced the short-player-buffer + disk-prefetch split)
- #238 (heap fix)
- #242 (timeline-aware prefetch planner — the spec this seems to violate)

## Affected file
- `app/src/main/java/org/hdhmc/saki/playback/SakiPlaybackService.kt` (`buildStreamPrefetchPlan`, `syncCurrentStreamPrefetch`, `SakiLoadControl` byte cap)

PengDingkang:
Closes #292.

## Problem
CUSTOM buffer streaming never reached the configured depth — upcoming tracks kept getting parked in the prefetch `deferred` state, well before the stream cache was anywhere near full (so not eviction).

## Root cause (device-traced — see #292)
`prefetchStreamToDisk` frequently returns **0 cached bytes with no exception**. That happens when the `CacheWriter` can't take the cache write-lock because the player is already caching that same track — the current item, or the next item being preloaded by the 10s `preloadConfiguration`. `prefetchTimelineToDisk` treated this harmless no-op as a failure and deferred the target for `STREAM_PREFETCH_RETRY_DELAY_MS` (30s).

Crucially, **real failures `throw`** and are handled by the job-level `runCatching`, so they never reach the deferral branch — the 30s backoff only ever penalized these transient contention no-ops. 30s was long enough to miss the brief window after the player releases the lock, so the next track stayed stuck with just the ~10s preload instead of the configured buffer.

## Fix
Shorten the no-op backoff from 30s to **3s** so the prefetch retries soon and fully caches the upcoming track once the lock releases. (Skipping the next track as a prefetch target was rejected — it would leave it with only the 10s preload.)

## Verification
On device with diagnostic logging: with the short retry, the next track (while still next) prefetches the **full file ahead (~26 MB)** and the plan reaches `done` for current + next, with no redownload churn (single result then quiet). Byte-cap was ruled out separately — the in-memory player buffer already fills the entire current track.

`./gradlew assembleDebug testDebugUnitTest` green.

### DIFF
--- a/app/src/main/java/org/hdhmc/saki/playback/SakiPlaybackService.kt
+++ b/app/src/main/java/org/hdhmc/saki/playback/SakiPlaybackService.kt
@@ -78,7 +78,13 @@ import okhttp3.OkHttpClient
 private const val CUSTOM_PLAYER_MAX_BUFFER_MS = 5 * 60 * 1_000
 private const val STREAM_PREFETCH_LOG_TAG = "SakiStreamPrefetch"
 private const val STREAM_PREFETCH_REEVALUATION_MS = 30_000L
-private const val STREAM_PREFETCH_RETRY_DELAY_MS = 30_000L
+// Backoff after a prefetch attempt cached 0 new bytes. This is only reached when the CacheWriter
+// returned without error but wrote nothing — i.e. the resource is momentarily locked by the player
+// caching the same track (the current item, or the next item being preloaded). Real failures throw
+// and are handled at the job level, so they never hit this path. Keep this short: the lock is held
+// only briefly, so a quick retry lets the prefetch fully cache the next track once it releases. A
+// long delay made the next track stall with just the ~10s preload instead of the configured buffer.
+private const val STREAM_PREFETCH_RETRY_DELAY_MS = 3_000L
 
 private fun Long.coerceKnownDuration(): Long? {
     return takeIf { it != C.TIME_UNSET && it > 0L }


PengDingkang:
Superseded by #294. Device tracing showed the retry/deferral length was not the root cause — the next track never fully caches ahead because the player's 10s next-item preload holds the same stream cache key and starves the disk prefetch. #294 fixes that by disabling preload only while disk prefetch is active. Closing this.

PR #308, opened 2026-06-27
fix: filter unknown album placeholders from feeds

### LINKED ISSUE #254: Reduce Unknown Album noise in album feeds
## Context

Album feeds can contain many pseudo albums named `[Unknown Album]` / `Unknown Album` when the server returns missing album metadata. The client also has a local fallback in `SubsonicMappers.kt` when an album DTO has neither `name` nor `title`.

In large libraries this can dominate the album tab and make sorted feeds / fast-scroll much less useful.

## Problem

We should not blindly remove these entries with an ad-hoc UI string filter, because songs without album metadata still need a reachable playback path and some servers may return pseudo albums with real IDs.

## Proposed direction

- Add an explicit unknown-album classification/normalization path in data/domain code, instead of scattering string checks in Compose UI.
- Decide whether default album feeds should hide pseudo unknown albums or collapse them into a single virtual "Unknown albums" entry.
- Preserve access to unalbumed songs through Songs/search/artist/offline cache, and through a virtual collection if hiding from album feeds would remove the only practical entry point.
- Treat the collapsed/unknown entry as a special `#` target in album fast-scroll instead of showing many repeated unknown album cards.

## Acceptance criteria

- Normal album feeds are no longer dominated by repeated `[Unknown Album]` entries.
- Songs with missing album metadata remain playable and discoverable.
- Real albums are not hidden just because their displayed name resembles an unknown placeholder.
- Behavior is consistent for server-returned placeholders, empty album names, and the client-side `Unknown Album` fallback.

PengDingkang:
## Summary
- Add a domain-level unknown-album placeholder helper for `Unknown Album` / `[Unknown Album]` names.
- Filter placeholder albums from album feed network results before displaying and caching them.
- Filter cached album feeds on read/write so old placeholder entries no longer keep polluting album feeds after refresh.
- Reuse the same placeholder detection in artist-detail album visibility logic.
- Keep unalbumed songs reachable through Songs/search/artist detail paths; this PR intentionally does not add a virtual Unknown Albums collection.

## Notes
- Pagination offsets continue to advance by the raw server page size, not the filtered visible count, so hidden placeholder rows do not cause repeated page requests.
- `loadMoreAlbums` keeps `hasMore` based on the raw server page size so filtered pages can still advance to later real albums.

## Validation
- `/usr/bin/git diff --check`
- `./gradlew testDebugUnitTest --tests org.hdhmc.saki.domain.model.UnknownAlbumTest --tests org.hdhmc.saki.domain.model.ArtistMatchingTest --no-daemon`
- `./gradlew compileDebugKotlin --no-daemon`

Closes #254

### DIFF
--- a/app/src/main/java/org/hdhmc/saki/data/repository/DefaultLibraryCacheRepository.kt
+++ b/app/src/main/java/org/hdhmc/saki/data/repository/DefaultLibraryCacheRepository.kt
@@ -37,6 +37,7 @@ import org.hdhmc.saki.domain.model.SearchResults
 import org.hdhmc.saki.domain.model.Song
 import org.hdhmc.saki.domain.model.visibleDetailAlbums
 import org.hdhmc.saki.domain.model.withVisibleDetailAlbums
+import org.hdhmc.saki.domain.model.withoutUnknownAlbumPlaceholders
 import org.hdhmc.saki.domain.repository.LibraryCacheRepository
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.withContext
@@ -91,11 +92,13 @@ class DefaultLibraryCacheRepository @Inject constructor(
     }
 
     override suspend fun getAlbums(serverId: Long, type: AlbumListType): List<AlbumSummary> = withContext(ioDispatcher) {
-        dao.getAlbums(serverId, type.apiValue).map { it.toDomain() }
+        dao.getAlbums(serverId, type.apiValue)
+            .map { it.toDomain() }
+            .withoutUnknownAlbumPlaceholders()
     }
 
     override suspend fun saveAlbums(serverId: Long, type: AlbumListType, albums: List<AlbumSummary>) = withContext(ioDispatcher) {
-        val entities = albums.mapIndexed { index, album ->
+        val entities = albums.withoutUnknownAlbumPlaceholders().mapIndexed { index, album ->
             CachedAlbumEntity(
                 serverId = serverId,
                 albumId = album.id,
--- a/app/src/main/java/org/hdhmc/saki/domain/model/ArtistMatching.kt
+++ b/app/src/main/java/org/hdhmc/saki/domain/model/ArtistMatching.kt
@@ -105,11 +105,11 @@ private fun AlbumSummary.hasPrimaryArtistMismatch(artist: Artist): Boolean {
 }
 
 private fun Album.hasUnknownAlbumName(): Boolean {
-    return name.isUnknownAlbumName()
+    return isUnknownAlbumPlaceholder()
 }
 
 private fun AlbumSummary.hasUnknownAlbumName(): Boolean {
-    return name.isUnknownAlbumName()
+    return isUnknownAlbumPlaceholder()
 }
 
 private fun Song.hasArtistIdentity(): Boolean {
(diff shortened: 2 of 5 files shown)


PengDingkang:
## Code Review

扎实、思路清晰的 PR。最棘手的部分(过滤 + 分页)处理得对。

**优点:**
- **分页计算在关键处是对的**:`loadMoreAlbums` 和网络加载路径都用**原始** `page.size` / `albums.size` 推进 `offset`,并以原始大小判定 `hasMore`,所以被隐藏的占位行不会导致重复请求同一窗口或提前判定到底。这正是"过滤+分页"的经典坑,这里避开了,PR 里那段说明也准确。
- `loadMoreAlbums` 去掉旧的 `addedAny` 守卫是 OK 的(其实更干净)—— 因为 `offset` 现在按原始大小推进,旧那套防无限加载的保护已不再需要,服务端最终会返回短页结束。
- 重构得当:`ArtistMatching.kt` 里重复的 `isUnknownAlbumName` 被共享的 `isUnknownAlbumPlaceholder()` 取代,归一化逻辑完全一致 → 那边行为无变化。
- `getAlbums()` **读取时过滤**是个好点 —— 能顺带迁移已经存了占位项的旧缓存,用户不必清缓存。
- 匹配器很精确:测试证明对 "Unknown Album Sessions" / "The Unknown Album" 这类真实名不会误杀。

**小问题(非阻塞):**
1. **缓存读取路径的 offset 不一致。** 它设 `offset = visibleCached.size`(过滤后),而网络路径用原始大小。正常情况下网络成功会覆盖它,但若网络刷新失败(或离线),随后的 `loadMore` 会从过滤后的计数开始 → 重新拉取有重叠的页(被 `distinctBy` 去重,所以只是冗余请求,不是错误数据)。由于原始服务端 offset 本来也没持久化,这算固有限制,但为一致性可以考虑对齐/注释一下。
2. **重复过滤:** 缓存读取路径里 VM 又调了一次 `cached?.withoutUnknownAlbumPlaceholders()`,但 `getAlbums()` 现在读取时已经过滤了。无害/防御性,提一下。
3. **基于名字的检测** 意味着用户若真有一张名字就叫 "Unknown Album" 的专辑,会被从 feed 隐藏。这是 #254 的有意取舍(歌曲仍可经 Songs/搜索/艺人详情到达),只是点出这个已知 tradeoff。
4. 测试对匹配器和列表过滤覆盖得不错;原始-offset 的分页逻辑(微妙的那块)因为内联在 VM 里没有单测 —— 如果将来把那段逻辑抽出来,可选地补一个。

小结:LGTM 带小建议 —— 无阻塞项。比 #306 干净很多。

PengDingkang:
Thanks, agreed on all points.

I’m leaving this PR as-is rather than expanding the scope:

- The cached-read `offset = visibleCached.size` mismatch is a known limitation because we do not persist the raw server offset for cached album feeds. In the offline/failed-refresh case it can cause a redundant overlapping page request, but `distinctBy` keeps the visible data correct.
- The extra ViewModel filter is intentionally defensive even though `DefaultLibraryCacheRepository.getAlbums()` now filters too.
- Hiding a real album named exactly `Unknown Album` / `[Unknown Album]` is the accepted tradeoff for #254; songs remain reachable through Songs/search/artist-detail paths.
- If we later extract album feed merge/pagination into a helper, I’ll add raw-offset tests there.

No further changes planned for this PR unless CI or manual verification finds an issue.

MYmeru (approved):

PR #318, opened 2026-06-29
fix: unify predictive back by surface role (#317)

### LINKED ISSUE #317: Unify predictive back behavior by surface role
## Problem

Predictive back support is currently implemented per surface, but the animation semantics are not yet consistently tied to the actual back target.

Recent queue-sheet testing exposed the mismatch clearly:

- Queue half-expanded -> expanded is an internal sheet anchor transition.
- Back from expanded currently should return to half-expanded, not close Now Playing or dismiss the whole queue.
- A page-exit style predictive back preview is misleading for that case.

This is a follow-up to #296. #296 established a clearer Browse detail page hierarchy; this issue tracks the broader app-level predictive back model.

## Goal

Define and implement predictive back behavior based on surface role, so every animated back gesture previews the same target that the completed back action will actually reach.

### Page / route surfaces
Examples:

- Now Playing full screen
- Settings
- Search overlay
- Browse detail pages

Animation direction:

- Use the existing page-surface predictive back treatment: scale, rounded corners, subtle translation, and optional fade.
- Reveal the real previous page/surface beneath.

### Bottom sheets / anchored surfaces
Examples:

- Queue sheet
- Other `ModalBottomSheet` surfaces

Animation direction:

- Do not use page-level scale/exit animation.
- Map predictive progress to sheet anchors:
  - `Expanded -> PartiallyExpanded`
  - `PartiallyExpanded -> Hidden`
- Scrim should follow the sheet dismiss/collapse state.

### Dialogs, menus, and transient UI
Examples:

- AlertDialog
- DropdownMenu
- Lightweight popups

Animation direction:

- Prefer Material/system defaults or a small fade/scale dismiss.
- These should not trigger page-level predictive back motion underneath.

### Internal state back
Examples:

- Queue expanded/fullscreen state
- Lyrics panel states
- Other in-page expanded/collapsed states

Animation direction:

- Back should undo the internal state first.
- Preview must match that state transition rather than looking like page dismissal.

## Suggested phases

1. Audit all current `BackHandler`, `PredictiveBackHandler`, `predictiveBackMotion`, `ModalBottomSheet`, dialog, and transient overlay usages.
2. Document the intended back target and animation type for each surface.
3. Keep `predictiveBackMotion` scoped to page/route surfaces.
4. Add a sheet-specific predictive back helper for anchored surfaces.
5. Apply it first to the queue sheet:
   - expanded back returns to half-expanded with sheet-anchor motion
   - half-expanded back dismisses the sheet
   - Now Playing page predictive back only runs when the queue sheet is not handling back
6. Revisit Search, Settings, Browse detail, Now Playing, dialogs, and other sheets for consistency.

## Acceptance criteria

- Back preview target always matches the completed back result.
- Queue expanded -> half-expanded uses sheet motion, not page-exit motion.
- Queue half-expanded -> hidden still dismisses as a sheet.
- Page/route surfaces keep page-style predictive back.
- Dialogs and menus do not accidentally trigger underlying page-scale predictive motion.
- Back cancellation does not flicker or leave surfaces in an intermediate scale/offset state.
- The behavior is documented enough that future surfaces can choose the correct animation role without ad-hoc fixes.

## Notes

A temporary local experiment can consume predictive back from the expanded queue state and call `partialExpand()` on commit, but that removes the live preview. The long-term fix should be a sheet-anchor predictive animation rather than simply suppressing the incorrect preview.

PengDingkang:
## Summary
Implements the surface-role predictive-back model from #317 so every animated back gesture previews the same target the completed back will reach.

### Phase 1 — route back to the topmost surface (Now Playing)
- Page-level `predictiveBackMotion` now runs only when no higher-priority surface is consuming back: `enabled = visible && !(showQueueSheet || showDetails || showMenu || showLyrics || showEndpointStatus)`. Stops the page-scale motion firing under dialogs/menus/lyrics/the queue sheet.
- Lyrics is an inline panel with no back handling, so back used to dismiss the whole page — added `BackHandler(enabled = visible && showLyrics) { showLyrics = false }` (internal-state back).

### Phase 2 — custom anchored queue sheet (no rebound)
- Replaced the M3 `ModalBottomSheet` queue with a custom `PlayerQueueSheet`. M3's built-in predictive back is dismissal-only and `SheetState` exposes no fractional offset, so it can't preview Expanded→PartiallyExpanded.
- Single `offsetY` float is the one source of truth for gesture + drag + nested scroll + animation. The predictive back **continues from the current offset** into the target anchor with a **critically-damped (`DampingRatioNoBouncy`) spring** — no hand-off and no bouncy overshoot, which removes the unnatural rebound on the Expanded→half back commit.
- M3-style settling: a flick past the velocity threshold (~125.dp/s) moves to the next anchor regardless of distance; slow drags fall back to nearest (positional). The whole header (handle + title) plus list nested-scroll is the drag area, so it expands one-handed.
- Scrim alpha + sheet translation are read inside `graphicsLayer{}` (deferred) so dragging/animating doesn't recompose the list.

### Server manager predictive back
- The server manager was the only page surface with functional back but no predictive animation. Added `predictiveBackMotion` (reveals Browse), gated `onCloseManager != null && editor == null` so forced first-run setup can't be swiped away. The editor sub-panel stays a transient surface (plain back dismiss, keeps its slide-out).

### Cleanup + docs
- Removed two redundant `BackHandler`s in `BrowseScreen` (detail pop / search close) that were always shadowed by the deeper `predictiveBackMotion` surfaces — zero behaviour change, removes a latent LIFO fragility.
- Documented the surface-role model in `PredictiveBackSurface.kt` (Page / Anchored sheet / Transient / Internal state) so future surfaces pick the right treatment.

## Surface-role audit (all conforming after this PR)
| Surface | Role | Treatment |
|---|---|---|
| Now Playing, Settings, Search, Browse detail, Server manager | Page | `predictiveBackMotion` (reveal previous) |
| Queue sheet | Anchored sheet | custom anchored predictive (Expanded↔Partial↔Hidden) |
| Dialogs, dropdown menu, single-anchor action sheet, server editor | Transient | default dismiss |
| Lyrics panel, queue expanded state | Internal state | back undoes the state first |
| Browse root | Root | system back exits |

## Validation
- `./gradlew assembleDebug --no-daemon` — BUILD SUCCESSFUL
- Needs on-device verification:
  - Queue: Expanded → back settles to half with **no rebound**; cancel eases back; half → back dismisses; one-handed flick-up expands; list scroll + drag-to-collapse.
  - Lyrics back closes the panel before exiting; dialogs/menus don't scale the page under them.
  - Server manager: edge-back reveals Browse; editor back closes editor first; first-run setup can't be swiped away.
  - Browse detail pop / search close still animate as before.

Closes #317

### DIFF
--- a/app/src/main/java/org/hdhmc/saki/presentation/PredictiveBackSurface.kt
+++ b/app/src/main/java/org/hdhmc/saki/presentation/PredictiveBackSurface.kt
@@ -23,6 +23,27 @@ import kotlinx.coroutines.CancellationException
 
 private val PredictiveBackInterpolator = PathInterpolator(0.1f, 0.1f, 0f, 1f)
 
+/**
+ * Predictive-back model by surface role (issue #317).
+ *
+ * Every animated back gesture must preview the same target the completed back will reach. Pick the
+ * treatment by what the surface *is*, not ad-hoc per screen:
+ *
+ * - **Page / route** (Now Playing, Settings, Search overlay, Browse detail pages): use
+ *   [predictiveBackMotion] — scale + rounded corners + slight translate, revealing the previous
+ *   surface. Gate `enabled` so it is OFF whenever a higher-priority surface is consuming back
+ *   (a sheet, dialog, menu, or an internal expanded state). See `NowPlayingOverlay`.
+ * - **Anchored sheet** (queue sheet): do NOT use page motion. Map predictive progress to the sheet
+ *   anchors (Expanded -> PartiallyExpanded -> Hidden) with a single offset source of truth and a
+ *   critically-damped (non-bouncy) commit so the settle never rebounds. Reference implementation:
+ *   `PlayerQueueSheet` in PlayerChrome.kt.
+ * - **Dialog / menu** (AlertDialog, DropdownMenu, single-anchor ModalBottomSheet): rely on the
+ *   Material/system default dismiss. These render in their own window / install their own back
+ *   handling, so they already pre-empt the page motion underneath — just make sure the page's
+ *   `predictiveBackMotion` is not *also* enabled for an inline (non-window) overlay.
+ * - **Internal state** (lyrics panel, other in-page expanded/collapsed states): a plain
+ *   `BackHandler` that undoes the state first, and keep the page motion disabled while it is shown.
+ */
 @Composable
 fun Modifier.predictiveBackMotion(
     enabled: Boolean,
--- a/app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
+++ b/app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
@@ -1,7 +1,6 @@
 package org.hdhmc.saki.presentation.library
 
 import android.icu.text.AlphabeticIndex
-import androidx.activity.compose.BackHandler
 import androidx.activity.compose.rememberLauncherForActivityResult
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.compose.foundation.BorderStroke
@@ -194,12 +193,9 @@ fun BrowseScreen(
     }
     val scrollState = rememberBrowseScrollState(uiState.selectedServerId)
 
-    BackHandler(enabled = backHandlersEnabled && uiState.browseStack.size > 1) {
-        onPopDetail()
-    }
-    BackHandler(enabled = backHandlersEnabled && uiState.browseStack.size <= 1 && uiState.isSearchActive) {
-        onSetSearchActive(false)
-    }
+    // Back for detail pop and search-close is owned by the page-level `predictiveBackMotion`
+    // surfaces below (the topmost detail page / the search overlay), which are composed deeper
+    // and so handle both the gesture and the back button with the reveal animation.
 
     Column(
         modifier = Modifier
--- a/app/src/main/java/org/hdhmc/saki/presentation/library/PlayerChrome.kt
+++ b/app/src/main/java/org/hdhmc/saki/presentation/library/PlayerChrome.kt
@@ -69,6 +69,7 @@ import androidx.compose.material.icons.rounded.SkipPrevious
 import androidx.compose.material3.AlertDialog
 import androidx.compose.material3.Button
 import androidx.compose.material3.Card
+import androidx.activity.compose.BackHandler
 import androidx.compose.material3.CardDefaults
 import androidx.compose.material3.DropdownMenu
 import androidx.compose.material3.DropdownMenuItem
@@ -465,6 +466,8 @@ fun NowPlayingOverlay(
     var showMenu by remember(track.songId) { mutableStateOf(false) }
     var showLyrics by remember { mutableStateOf(false) }
     var showEndpointStatus by remember { mutableStateOf(false) }
+    val queueSheetState = rememberBottomSheetState(initialValue = SheetValue.Hidden)
+    var showQueueSheet by remember { mutableStateOf(false) }
     val visualSnapshot = rememberNowPlayingVisualSnapshot(
         queue = playbackState.queue,
         currentIndex = playbackState.currentIndex,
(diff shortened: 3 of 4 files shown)


PengDingkang:
Review conclusion: I don’t think this PR should be merged as-is.

The general surface-role direction looks right, but the custom queue sheet implementation is currently the risky part and matches the on-device issues we already saw.

## Blocker 1: the queue sheet cannot cover the full visual surface

`PlayerQueueSheet` is composed inside the Now Playing content container that already applies:

```kotlin
.statusBarsPadding()
.navigationBarsPadding()
.imePadding()
```

So the sheet’s `fillMaxSize()` / `fillMaxHeight()` only fills the padded content area, not the whole window. On top of that, the expanded anchor is hardcoded to:

```kotlin
val expandedOffset = 56.dp.toPx()
```

That means the “expanded/fullscreen” state intentionally still leaves 56dp exposed at the top. This explains the observed top/bottom leakage of the Now Playing background. This is not just a tuning issue; fixing it would require moving the custom sheet to a non-padded top-level overlay/window layer and handling insets manually.

## Blocker 2: predictive back does not actually start from the current offset

The PR body says the predictive back continues from the current offset, but the implementation uses the settled anchor as the preview start:

```kotlin
val start = settledAnchor
val from = offsetFor(start)
```

`settleTo()` also updates `settledAnchor` before the animation reaches that anchor:

```kotlin
settledAnchor = anchor
```

So during opening/settling, after a partial drag, or after a cancelled motion, `settledAnchor` can diverge from the real `offsetY`. Starting predictive back from `offsetFor(settledAnchor)` can then jump/flicker instead of continuing from the actual current position.

## What looks OK

- Gating Now Playing page-level predictive back while queue/dialog/menu/lyrics surfaces are active is the right direction.
- Adding lyrics as an internal-state back handler is reasonable.
- Server manager as a page-level predictive-back surface looks reasonable.
- Removing the redundant Browse back handlers is acceptable because detail/search still have deeper page-level predictive surfaces.

## Suggested path

I’d revert the custom `PlayerQueueSheet` replacement and restore `ModalBottomSheet` for this PR. Keep the safer surface-role changes:

- Now Playing page-level back gating.
- Lyrics internal-state back.
- Server manager predictive back.
- Predictive-back surface-role docs.

Queue expanded ↔ half ↔ hidden predictive back should stay in #317 as follow-up work. The custom sheet needs a cleaner top-level/inset architecture before it is worth merging.

PengDingkang:
两个 Blocker 都修了(并在设备上验证了),没有退回 ModalBottomSheet。新增提交 `7d02b66`。

## Blocker 1 — sheet/scrim 没盖满整窗
`PlayerQueueSheet` 之前在带 `.statusBarsPadding().navigationBarsPadding().imePadding()` 的 Now Playing `BoxWithConstraints` 内部,所以 scrim 从状态栏下方才开始。已把它移到 **Now Playing 覆盖层根部**(和那几个 `AlertDialog` 同一层,padding 之前),现在 scrim `fillMaxSize` 覆盖含状态栏的整窗;`verticalSpacing` 改为 sheet 用自身约束内部计算(不再依赖那个 padded 容器里的值)。

设备验证:展开态顶部小幅露出(可接受)、底部不漏、**状态栏沉浸区现在也被遮罩覆盖**。

## Blocker 2 — predictive 没从当前 offset 起步
确实如你所说:之前 `from = offsetFor(settledAnchor)`,而 `settledAnchor` 会在动画到位前就更新,导致起手可能跳变。已改为 `val from = offsetY`(捕获手势开始那一刻的真实偏移),不再跳。

## 顺带修掉的惯性 bug(同一类问题)
测试中发现 nested scroll 会把**列表 fling 的动量**也用来驱动 sheet,导致:(a) 列表 fling 未停时返回会回弹/重新展开;(b) 滚进列表后用力下滑会直接滑到关闭。已改为 `onPreScroll`/`onPostScroll` 仅在 `source == NestedScrollSource.UserInput`(用户主动拖动)时移动 sheet,**fling 动量一律不驱动 sheet**——fling 只滚列表,只有主动拖把手/列表顶部下拖才会收起/关闭。两个场景均已在设备上验证修复。

`assembleDebug` 通过 + 设备验证。你approve的那几项(Now Playing 门控 / 歌词内部状态返回 / 服务器管理 page 预测返回 / 删冗余 Browse handler / 角色文档)不变。麻烦再看一下,谢谢。

MYmeru (approved):

PR #320, opened 2026-06-30
fix(artist): reconcile split-artist album/song associations + opt-in hide merged artists

PengDingkang:
## Problem
When the music server (Navidrome/OpenSubsonic) attributes an **album** to a combined/composite artist while crediting the album's **tracks** to the individual (split) artists, the app's artist views diverged:

- The combined artist (e.g. `23.exe/初音ミク`) showed the album but **no songs**.
- Each split artist (e.g. `23.exe`) showed songs but **no album**.

This also covered spelling variants where the album-artist id differs from the track-artist id (e.g. `安田 レイ` with a space vs `安田レイ` without).

Root cause: the **main** `getArtistDetail` branch (used once an artist detail row is cached) took albums only from the server's artist-detail albums, while songs came from the per-song artist relationships (`cached_song_artists`). The two never reconciled. The network refresh in `openArtist` compounded it by rebuilding the artist from the server response and dropping any relationship-derived albums.

All detection here is **relationship-only and separator-agnostic** — it relies solely on the server-provided per-song artist rows, never on parsing artist names (the set of separators is configured server-side).

## Changes

**1 — Associate albums with split artists via song relationships**
- New DAO query `getArtistAlbumRefsFromSongs` derives `(artist, album)` pairs from `cached_song_artists` ⋈ `cached_song_metadata` (excluding `[Unknown Album]` placeholders).
- `getArtistDetail` (main branch) now unions server detail albums with albums derived from the artist's related songs.
- `openArtist`'s network refresh now merges relationship-derived albums into the displayed artist so they aren't dropped after the network result arrives.

**2 — Artist-list album count parity**
- The list album count now mirrors the detail page: visible detail albums ∪ relationship-derived albums that are absent from the detail set, with unknown-album placeholders excluded so the list no longer over-counts vs the detail (e.g. 初音ミク showed 101 in the list but 99 in the detail).

**3 — Opt-in "try to hide merged artists" toggle** (new switch in the song-metadata settings card)
- Hides server-side composite/variant album-artists that **no track credits** yet that **claim albums** — their albums remain reachable through the artists the tracks actually credit (via change 1).
- Decided purely from list-level data (`cached_song_artists` + the index's `albumCount`), so the result is identical before and after opening an artist (an earlier approach that relied on lazily-cached detail albums was inconsistent). No-op when there is no song relationship data. Best run after "Update all metadata".

## Testing
- `assembleDebug` + `testDebugUnitTest` pass.
- Verified on device against the live database:
  - `23.exe` and `初音ミク` now both show the `LIME` album.
  - Toggle hides `23.exe/初音ミク` and `安田 レイ` from the artist list (consistently, without needing to open them); off restores them.
  - 初音ミク list count now matches the detail (99), with the two `[Unknown Album]` placeholders correctly excluded.

### DIFF
--- a/app/src/main/java/org/hdhmc/saki/data/local/dao/LibraryCacheDao.kt
+++ b/app/src/main/java/org/hdhmc/saki/data/local/dao/LibraryCacheDao.kt
@@ -126,6 +126,10 @@ interface LibraryCacheDao {
         JOIN cached_song_metadata AS song
             ON song.serverId = artist.serverId AND song.songId = artist.songId
         WHERE artist.serverId = :serverId AND song.albumId IS NOT NULL
+            AND (
+                song.album IS NULL
+                OR LOWER(TRIM(REPLACE(REPLACE(song.album, '[', ''), ']', ''))) != 'unknown album'
+            )
         """,
     )
     suspend fun getArtistAlbumRefsFromSongs(serverId: Long): List<ArtistAlbumRef>
--- a/app/src/main/java/org/hdhmc/saki/data/local/dao/LibraryCacheDao.kt
+++ b/app/src/main/java/org/hdhmc/saki/data/local/dao/LibraryCacheDao.kt
@@ -19,6 +19,7 @@ import org.hdhmc.saki.data.local.entity.CachedPlaylistDetailSongEntity
 import org.hdhmc.saki.data.local.entity.CachedPlaylistEntity
 import org.hdhmc.saki.data.local.entity.CachedSongMetadataEntity
 import org.hdhmc.saki.data.local.entity.CachedSongMetadataOrder
+import org.hdhmc.saki.data.local.entity.ArtistAlbumRef
 import org.hdhmc.saki.data.local.entity.CachedSongArtistEntity
 import org.hdhmc.saki.data.local.entity.CachedSongArtistSummary
 import kotlinx.coroutines.flow.Flow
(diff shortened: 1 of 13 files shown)


PengDingkang:
Review conclusion: direction looks right, but I’d fix one issue before merging.

## Issue: `hideMergedArtists` does not filter `shortcuts`

The hide logic currently only applies to the merged artist list used for `sections`:

```kotlin
val artists = mergeArtistSummaries(...).map { ... }.let { merged ->
    if (!hideMergedArtists) return@let merged
    ...
    merged.filterNot { artist ->
        artist.id !in claimedArtistIds && (artist.albumCount ?: 0) > 0
    }
}

LibraryIndexes(
    shortcuts = shortcutEntities.map { it.toDomain() },
    sections = listOf(ArtistSection(name = "#", artists = artists)),
)
```

But the artist UI renders `indexes.shortcuts` separately at the top:

```kotlin
if (indexes.shortcuts.isNotEmpty()) {
    LazyRow {
        items(indexes.shortcuts, key = { it.id }) { artist ->
            ArtistShortcutCard(...)
        }
    }
}
```

`regroupByLocale()` only regroups `sections`; it does not remove or rewrite shortcuts. So a composite/variant artist hidden from the main list can still remain visible in the shortcuts row if the server provides it there.

Suggested fix: compute a shared hidden-artist predicate / hidden id set and apply it to both:

- `sections` artists
- `shortcuts`

Everything else looks reasonable:

- Relationship-derived `(artist, album)` association is separator-agnostic and does not parse names.
- `getArtistDetail` unioning server/detail albums with relationship-derived albums solves the split-artist “songs but no album” case.
- `openArtist` preserving relationship-derived albums after network refresh is necessary.
- The album-count parity logic is defensible.
- The setting is opt-in and the “Try to hide” wording is appropriately conservative.

CI build was still running when I checked; I’d fix shortcuts filtering and wait for CI before merge.

PengDingkang:
Good catch — fixed in `def81df`.

Hoisted the hide logic into a shared `isMergedArtist(id, albumCount)` predicate and applied it to **both** the `sections` artists and the `shortcuts` row, so a hidden composite/variant artist can no longer reappear in the shortcuts strip. (`regroupByLocale()` only touches `sections`, as you noted, so the filtering has to happen here before the `LibraryIndexes` is built.)

The predicate keeps the same semantics: only active when `hideMergedArtists` is on and there is at least some song-relationship data, hiding artists that no track credits but that still claim albums.

`assembleDebug` + `testDebugUnitTest` pass; verified on device that the hidden artists no longer show in the shortcuts row either. Ready for re-review once CI is green.

MYmeru (approved):

PR #321, opened 2026-06-30
feat(navigation): add enter animations for page/route surfaces

PengDingkang:
## Summary
- Add `Modifier.pageEnterMotion()` — a crisp forward "cover" open (rise ~6% + fade in, ~250ms) for page/route surfaces; applied to Browse detail pages, the search overlay, the server manager, and Settings (replacing its plain fade so all page opens match).
- Refactor `predictiveBackMotion` into a shared `PredictiveBackMotionState` (`rememberPredictiveBackMotion`) exposing both the `modifier` and a `dismiss()`; keep `Modifier.predictiveBackMotion()` as a thin wrapper for gesture-only call sites.
- Make in-app back/close buttons animate via `dismiss()`: a tap has no gesture to track, so it plays a quick reverse-of-open (slide down + fade out, ~160ms) rather than the predictive follow-finger commit. Wired through Browse detail pages, the Settings close button (restoring an animated close), and the search overlay close.
- Leave already-animated roles unchanged: Now Playing (rise from the bottom capsule), anchored sheets (queue / song actions), dialogs / menus (system default), and the server-manager close (system back only, already animated).

## Validation
- On-device: detail / search / server-manager / settings open with the rise-and-fade; back arrows and close buttons animate with a fast slide-down + fade; predictive gesture and system back are unchanged.

### DIFF
--- a/app/src/main/java/org/hdhmc/saki/presentation/PredictiveBackSurface.kt
+++ b/app/src/main/java/org/hdhmc/saki/presentation/PredictiveBackSurface.kt
@@ -81,6 +81,13 @@ fun Modifier.predictiveBackMotion(
  * button animates identically to the gesture (and to the system back button, which
  * [PredictiveBackHandler] already animates).
  */
+/**
+ * Holds the motion state for a page/route surface and exposes both the [modifier] that renders it
+ * and a programmatic [dismiss]. The predictive gesture — and the system back button, which
+ * [PredictiveBackHandler] drives — animate the follow-the-finger preview + commit. An in-app
+ * back/close button should instead go through [dismiss], which plays a quick reverse-of-open
+ * (slide down + fade out) before invoking `onBack`.
+ */
 @Stable
 class PredictiveBackMotionState internal constructor(
     private val scope: CoroutineScope,
@@ -98,18 +105,26 @@ class PredictiveBackMotionState internal constructor(
 ) {
     internal var swipeEdge by mutableIntStateOf(BackEventCompat.EDGE_LEFT)
     internal var touchY by mutableFloatStateOf(0f)
+    private var isDismissing = false
 
     /**
      * Programmatic close for an in-app back/close button. This is NOT the predictive follow-finger
      * commit — a tap has no gesture to track. Instead it plays a quick reverse-of-open (slide down
-     * + fade out), then invokes onBack.
+     * + fade out), then invokes onBack. Ignores re-entrant calls so a double-tap can't invoke
+     * onBack more than once.
      */
     fun dismiss() {
+        if (isDismissing) return
+        isDismissing = true
         scope.launch {
-            if (dismissProgress.value < 1f) {
-                dismissProgress.animateTo(1f, animationSpec = spring(dampingRatio = 1f, stiffness = 900f))
+            try {
+                if (dismissProgress.value < 1f) {
+                    dismissProgress.animateTo(1f, animationSpec = spring(dampingRatio = 1f, stiffness = 900f))
+                }
+                onBack()
+            } finally {
+                isDismissing = false
             }
-            onBack()
         }
     }
 
--- a/app/src/main/java/org/hdhmc/saki/presentation/PredictiveBackSurface.kt
+++ b/app/src/main/java/org/hdhmc/saki/presentation/PredictiveBackSurface.kt
@@ -109,3 +109,36 @@ fun Modifier.predictiveBackMotion(
         clip = true
     }
 }
+
+/**
+ * Enter motion for a hierarchical child **page/route** (e.g. a Browse detail pushed onto the
+ * stack). Open is a discrete action, not a gesture, so this is NOT the inverse of the back
+ * preview — it is a crisp, confident forward "cover": the incoming page rises a short distance
+ * and fades in over its parent, then settles. Pairs coherently with the shrink-away back without
+ * trying to rewind it.
+ *
+ * The animation runs once per composition; wrap the page in a `key(route)` (as Browse does) so a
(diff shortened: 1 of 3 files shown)


PengDingkang:
Review note: overall direction looks good, but one behavior regressed for Settings.

## Settings non-gesture close now snaps away

This PR replaces the previous Settings wrapper:

```kotlin
AnimatedVisibility(
    visible = showSettings,
    enter = fadeIn(...),
    exit = fadeOut(...),
) { ... }
```

with:

```kotlin
if (showSettings) {
    Box(
        modifier = Modifier
            .fillMaxSize()
            .pageEnterMotion()
            .then(settingsBackModifier)
    ) { ... }
}
```

This gives Settings the new enter motion, but it removes the old exit transition for non-gesture closes. The close button path still does:

```kotlin
onClose = { onShowSettingsChange(false) }
```

so Settings is immediately removed from composition and disappears instantly. Predictive back itself is OK because `predictiveBackMotion` runs its commit before calling `onBack`, but the explicit close button / programmatic close behavior is no longer equivalent to before. That conflicts with the PR body saying back/close behavior is unchanged.

Suggested fix: keep an exit transition around Settings while still using `pageEnterMotion()` for enter. For example, conceptually:

```kotlin
AnimatedVisibility(
    visible = showSettings,
    enter = EnterTransition.None,
    exit = fadeOut(...),
) {
    Box(
        modifier = Modifier
            .fillMaxSize()
            .pageEnterMotion()
            .then(settingsBackModifier)
    ) { ... }
}
```

The exact spec can be tuned, but we should avoid making the close button snap the page away.

Everything else looks reasonable: `pageEnterMotion()` is simple, Browse detail/search enter motion fits the page-role model, and it correctly avoids applying this to Now Playing, sheets, dialogs, and menus.

PengDingkang:
Fixed in `7010045`, and went a step further to address the general case.

## Settings non-gesture close
You were right — replacing the `AnimatedVisibility(exit = fadeOut)` with a bare `if (showSettings)` dropped the animated close for the button/programmatic path. Rather than re-adding a separate `AnimatedVisibility` exit (which would double-animate against the predictive commit on the gesture path), the close button now animates through a shared back-motion state.

## General fix: in-app back/close buttons
The same gap existed on every page/route surface: the predictive gesture and the system back button animate, but in-app back/close buttons called `onBack`/`onClose` directly and removed the surface instantly.

- Refactored `predictiveBackMotion` into a shared `PredictiveBackMotionState` (`rememberPredictiveBackMotion`) exposing both the `modifier` and a `dismiss()`.
- A tap has no gesture to track, so `dismiss()` deliberately does **not** replay the predictive follow-finger commit; it plays a quick **reverse-of-open** (slide down + fade out, critically damped ~160ms) then invokes `onBack`. Gesture progress and dismiss progress are independent tracks.
- Wired through `dismiss()`: Browse detail pages (per-page state), Settings close, search overlay close. Server manager closes only via the system back button (already animated), so it's unchanged. `Modifier.predictiveBackMotion()` is kept as a thin wrapper for gesture-only sites.

`assembleDebug` + `testDebugUnitTest` pass; verified on device that back arrows / close buttons now animate (fast slide-down + fade) while the gesture/system-back behavior is unchanged.

PengDingkang:
Re-review: the Settings close regression is fixed, and the generalized `rememberPredictiveBackMotion` direction is reasonable. I’d still fix two small issues before merge.

## 1. Guard against repeated `dismiss()` calls

Current `dismiss()` can launch multiple concurrent dismiss coroutines:

```kotlin
fun dismiss() {
    scope.launch {
        if (dismissProgress.value < 1f) {
            dismissProgress.animateTo(1f, animationSpec = spring(dampingRatio = 1f, stiffness = 900f))
        }
        onBack()
    }
}
```

If the user double-taps an in-app back/close button before the first animation finishes, `onBack()` can run more than once.

Risk examples:
- Browse detail back arrow double-tap may pop two detail levels.
- Search/settings are lower risk, but still issue redundant state changes.

Suggested fix: add an `isDismissing` guard or keep a dismiss job so repeated calls while a dismiss is in flight are ignored.

Conceptually:

```kotlin
if (isDismissing) return
isDismissing = true
scope.launch {
    try {
        dismissProgress.animateTo(1f, animationSpec = ...)
        onBack()
    } finally {
        isDismissing = false
    }
}
```

The surface will usually leave composition after `onBack()`, so the important part is preventing concurrent launches before that happens.

## 2. KDoc wording is now stale

`PredictiveBackMotionState` KDoc still says `dismiss()` “replays the same commit animation a completed back gesture would”, but the implementation and PR body correctly say the button path is a reverse-of-open slide/fade, not the predictive gesture commit.

Please adjust that KDoc to match the final behavior so future changes do not assume `dismiss()` and gesture commit are the same animation.

After those two small fixes and CI green, this looks mergeable.

PengDingkang:
Both addressed in `abef5d8`.

1. **Re-entrant `dismiss()`** — added an `isDismissing` guard: the first call sets it, the coroutine clears it in a `finally`, and concurrent calls return early. So a double-tap on a back/close button can no longer invoke `onBack()` twice (e.g. pop two detail levels).
2. **Stale KDoc** — updated the `PredictiveBackMotionState` class doc to state that `dismiss()` plays a reverse-of-open (slide down + fade out), distinct from the gesture/system-back follow-finger commit.

`assembleDebug` + `testDebugUnitTest` pass.

PengDingkang:
Good catch — my earlier "fixed" edit prepended the corrected KDoc but left the old one, so there were two contradictory blocks. Removed the stale block in `58a1f29`; only the correct KDoc (`dismiss()` = reverse-of-open slide/fade) remains.

MYmeru (approved):

PR #322, opened 2026-06-30
fix(server-config): explicit close affordance, icon close button, editor exit animation

PengDingkang:
## Summary
- Add an explicit back affordance to the server manager page (a `SakiChromeIconButton` with `ArrowBack`, matching Settings), shown only when the manager is closeable so the forced first-run setup can't be dismissed. It routes through the page's animated `dismiss()`, so the button, the predictive gesture, and the system back button all animate consistently.
- Replace the editor (add/edit) panel's text "Close" `TextButton` with the `Close` (X) `SakiChromeIconButton`, so the close/back affordance matches the arrow/X icons used elsewhere instead of text.
- Fix the editor panel enter/exit animation: the `AnimatedVisibility` content read `uiState.editor` directly, so on dismiss it was already null and there was nothing to slide out (the slide-out never played). Retain the last editor (`lastEditor`) for the duration of the exit so the slide-down + fade-out actually animates.

## Validation
- On-device: the manager shows a back arrow that animates closed (and the predictive gesture/system back still work); the editor opens with a slide-up and closes (X or back) with a slide-down + fade.

## Notes
The editor remains a transient bottom-sheet surface (slide in/out, plain back dismiss) per the #317 surface-role model, rather than adopting page-style predictive back.

### DIFF
--- a/app/src/main/java/org/hdhmc/saki/presentation/serverconfig/ServerConfigScreen.kt
+++ b/app/src/main/java/org/hdhmc/saki/presentation/serverconfig/ServerConfigScreen.kt
@@ -118,9 +118,15 @@ fun ServerConfigRoute(
     ServerConfigScreen(
         uiState = uiState,
         snackbarHostState = snackbarHostState,
-        // Explicit back affordance (arrow) shares the same animated commit as the gesture; only
-        // shown when the manager is closeable (i.e. not the forced first-run setup).
-        onClose = onCloseManager?.let { managerBackMotion::dismiss },
+        // Explicit back affordance (arrow). Plays the in-app reverse-of-open close animation via
+        // dismiss() (not the predictive gesture commit). Gated exactly like the predictive-back
+        // handler: hidden during the forced first-run setup, and while the editor is open so the
+        // editor consumes back first.
+        onClose = if (onCloseManager != null && uiState.editor == null) {
+            managerBackMotion::dismiss
+        } else {
+            null
+        },
         modifier = modifier.then(managerBackMotion.modifier),
         onAddServer = viewModel::startAddingServer,
         onEditServer = viewModel::editServer,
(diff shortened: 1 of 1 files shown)


PengDingkang:
Review note: this looks correct overall, but I’d fix two small consistency issues before merge.

1. Gate the explicit manager close affordance the same way as the predictive-back handler. The motion is disabled while the editor is open:

```kotlin
enabled = onCloseManager != null && uiState.editor == null
```

but the UI callback is still present:

```kotlin
onClose = onCloseManager?.let { managerBackMotion::dismiss }
```

The scrim likely prevents normal pointer access, but semantically/testing/accessibility-wise this can still expose a manager close path while the editor should consume back first. Prefer:

```kotlin
onClose = if (onCloseManager != null && uiState.editor == null) {
    managerBackMotion::dismiss
} else {
    null
}
```

2. The comment says the arrow “shares the same animated commit as the gesture”, but after #321 `dismiss()` intentionally uses the in-app button reverse-of-open animation, not the predictive gesture commit. Please update that wording so the surface-role/back-animation model stays unambiguous.

No broader blocking concerns in the diff; wait for CI after this cleanup.

PengDingkang:
Addressed in `bcb4546` + `607ab5f`.

1/2. Gating `onClose` to `null` while the editor is open (your snippet) worked semantically but caused a visible regression: the back arrow's visibility is tied to `onClose != null`, so opening the editor removed the arrow, shrank the manager's first list item, and reflowed the whole page up into the status bar (a visible jump on "add server").

Kept your intent without the reflow: the arrow's **visibility** is now stable (shown whenever the manager is closeable, i.e. `onCloseManager != null`), while the **action** is gated — `onClose = onCloseManager?.let { { if (uiState.editor == null) managerBackMotion.dismiss() } }`. So while the editor is open the arrow does nothing (the editor consumes back first), matching the predictive-back handler, and the page no longer reflows.

Also corrected the comment to say `dismiss()` is the in-app reverse-of-open close, not the predictive gesture commit.

`compileDebugKotlin` passes.

MYmeru (approved):

PR #323, opened 2026-07-01
feat(theme): system dynamic color option with palette-style support and pokeball swatches

PengDingkang:
## Summary
- Add a **System** (Material You) option to the Material Expressive theme-color picker, shown only on Android 12+. Selecting it sources the app's colors from the system/wallpaper palette.
- Route the system color through the same MaterialKolor seed pipeline as the presets (seeded from `system_accent1_500`), so all three palette styles (Tonal Spot / Vibrant / Expressive) still apply and the app's calm desaturation stays consistent — previously the platform scheme ignored palette style.
- Redesign the theme swatches as KernelSU-style "pokéball" tiles (scaled down for Settings): a `surfaceContainer` tile with a two-tone ball (top = `primaryContainer`, bottom = `secondaryContainer`) and a `primary` center button; selection shows a ringed check. Roles chosen to match what the app actually renders (dropped `tertiaryContainer`, which is effectively unused). The system swatch shows the phone marker in its center button.
- Fix a scroll frame drop: the theme swatch colors (8 generated schemes) were computed the moment the Settings theme item composed mid-scroll and recomputed on every recycle. They are now precomputed once in a hoisted `remember` keyed on `(dark, paletteStyle, themeStyle)`, off the scroll path.

## Validation
- On-device: Material Expressive → theme color → System follows the wallpaper; Tonal Spot / Vibrant / Expressive each render distinctly under System; scrolling to the theme section is smooth; picker unavailable below Android 12.

## Notes
The default **Saki** style already used system dynamic color on Android 12+; this exposes the choice explicitly for the Expressive style, consistent with that style being the only one with a color picker.

### DIFF
--- a/app/src/main/java/org/hdhmc/saki/presentation/SakiApp.kt
+++ b/app/src/main/java/org/hdhmc/saki/presentation/SakiApp.kt
@@ -44,6 +44,7 @@ import org.hdhmc.saki.presentation.library.NowPlayingOverlay
 import org.hdhmc.saki.presentation.serverconfig.ServerConfigRoute
 import org.hdhmc.saki.presentation.settings.SettingsScreen
 import org.hdhmc.saki.ui.theme.seedColorForKey
+import org.hdhmc.saki.ui.theme.isSystemDynamicSeed
 import org.hdhmc.saki.ui.theme.SakiAndroidTheme
 
 @Composable
@@ -100,6 +101,7 @@ fun SakiApp(
         themeStyle = rootUiState.appPreferences.themeStyle,
         seedColor = seedColorForKey(rootUiState.appPreferences.themeSeedKey),
         paletteStyle = rootUiState.appPreferences.paletteStyle,
+        useSystemColor = isSystemDynamicSeed(rootUiState.appPreferences.themeSeedKey),
     ) {
         CompositionLocalProvider(LocalDensity provides appDensity) {
         Box(modifier = modifier.fillMaxSize()) {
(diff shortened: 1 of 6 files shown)


PengDingkang:
Review note: implementation direction looks good, but I’d fix the swatch semantics before merge.

The old seed picker used selectable radio semantics:

```kotlin
.selectable(
    selected = selected,
    role = Role.RadioButton,
    onClick = { onUpdateThemeSeed(preset.key) },
)
```

The new `ThemeSeedSwatch` is visually selected via the check/ring, but `Surface(onClick = ...)` makes it semantically closer to a plain button. Since this is a mutually-exclusive theme picker, please preserve the RadioButton/selected semantics for TalkBack and UI tests — either by using `Modifier.selectable(selected = isSelected, role = Role.RadioButton, ...)` around the tile, or by adding equivalent selectable semantics while avoiding a duplicate click target.

Minor optional edge case: below Android 12, the System swatch is hidden. If a backed-up/restored preference already contains `"system"`, no preset appears selected even though the actual seed falls back to HarborBlue. Not blocking, but worth normalizing or displaying the fallback as selected if this is easy.

No broader concerns from the diff; wait for CI after the semantics cleanup.

PengDingkang:
Both addressed in `d5621d7`.

1. Restored radio semantics: the tile now uses `Modifier.selectable(selected = isSelected, role = Role.RadioButton, onClick = onClick)` (with `.clip(shape)` so the ripple stays within the rounded tile) instead of `Surface(onClick = ...)`, so it's a single click target that announces as a selected/unselected radio button for TalkBack and UI tests.
2. Below Android 12, an `effectiveSeedKey` now maps a stored `"system"` seed to `DEFAULT_THEME_SEED_KEY` (HarborBlue) for the selection state, so the brand preset shows selected and matches the actual `seedColorForKey` fallback instead of leaving nothing highlighted.

`assembleDebug` + `testDebugUnitTest` pass.

MYmeru (approved):

Issue #324, opened 2026-07-02
Adaptive navigation and large-screen Browse layout


PengDingkang:
Parent: #251

## Context

#251 still has one major unresolved area: the app does not yet adapt root navigation or Browse content layout for medium / expanded window sizes. The phone experience has received most M3E polish, but tablets, foldables, desktop-sized windows, and split-screen still mostly reuse the compact phone structure.

## Goal

Introduce adaptive navigation and large-screen-aware Browse layout incrementally, without destabilizing the compact phone UI.

## Tasks

1. Add Material 3 adaptive navigation dependencies and migrate the root app navigation chrome to a `NavigationSuiteScaffold`-style shell, keeping the current compact bottom/tab behavior on phones.
2. Add window-size-aware Browse album grids, replacing fixed two-column album grids with adaptive columns for medium/expanded widths.
3. Evaluate whether library detail pages should stay single-pane or adopt a supporting/list-detail pane on expanded widths.
4. Verify Settings, Server Manager, Now Playing, and Search overlays still layer correctly over the adaptive shell.

## Acceptance criteria

- Compact phone layout remains visually close to current behavior.
- Medium/expanded widths use navigation chrome appropriate for the available width instead of only phone-style tab/chip navigation.
- Album grids adapt beyond compact width.
- Settings, server management, search, and Now Playing still open/close correctly over the shell.
- `./gradlew assembleDebug` succeeds.

## Out of scope

- Reworking every detail page into full tablet layouts in the first PR.
- Playback, queue, cache, or Subsonic data behavior changes.
- Visual redesign unrelated to adaptive navigation.

Related: #251

Issue #251, opened 2026-05-25
Track UI/UX and Material 3 Expressive alignment gaps


PengDingkang:
## Problem

The README describes Saki.Android as a Material 3 Expressive music player, but the current UI implementation is closer to a custom Material 3 skin. A static pass over the Compose UI found several places where the app can drift from M3/M3E expectations or produce weak UX on accessibility, large screens, and dynamic color.

Main gaps:

- Major app destinations are driven by chip rows and nested horizontal pagers instead of adaptive navigation (`NavigationSuiteScaffold`, rail/bar/drawer based on window size).
- Large-screen and foldable layouts are not differentiated; album grids are fixed at two columns and server editor fields force two-column rows on compact screens.
- The mini player transport controls expose icon-only buttons with `contentDescription = null`.
- The Settings screen is shown as a full-screen overlay without a visible close/back affordance.
- Custom alpha surfaces over gradient backgrounds can weaken contrast under dynamic color.
- The custom alphabet fast scroller is pointer-only and has no accessibility semantics.
- The app uses normal Material 3 components/theme APIs, but does not currently use M3E-specific components or expressive APIs such as `LoadingIndicator`, `ButtonGroup`, or expressive color/theme helpers.

## Expected behavior

- Phone, tablet, foldable, and desktop-style window sizes should receive appropriate navigation and content layouts.
- Icon-only interactive controls should be named for TalkBack and other assistive tech.
- Settings and modal surfaces should expose clear visible exit affordances.
- Color and surface treatment should preserve contrast across dynamic color, light/dark theme, and gradient backgrounds.
- The implementation should either adopt M3E APIs where they materially improve the UI or soften the README claim to "Material 3".

## Proposed fix

1. Add Material 3 adaptive dependencies and migrate root navigation to `NavigationSuiteScaffold`.
2. Introduce window-size-aware browse layouts:
   - compact: current single-column phone layout
   - medium/expanded: wider album grids, optional list-detail/supporting-pane layouts for library detail
3. Fix accessibility labels for `NowPlayingCapsule` transport buttons and add semantics/custom actions for `AlphabetScrollBar`.
4. Add a visible close/back action to Settings, or make Settings a real navigation destination with selected state.
5. Replace arbitrary translucent `surface.copy(alpha = ...)` containers with `surfaceContainer*`/tonal roles where possible, then run contrast checks under dynamic colors.
[...]

PR #325, opened 2026-07-02
[codex] Add adaptive Browse navigation

PengDingkang:
## Summary
- Add the Material 3 adaptive navigation suite dependency.
- Use `NavigationSuiteScaffold` with a left `NavigationRail` for Browse section navigation on widths >= 600dp.
- Keep the existing compact phone chip navigation below 600dp so phone portrait behavior remains effectively unchanged.

## Scope
This implements the first task from #324. Album grid column adaptation and detail-pane layouts remain follow-up work.

## Validation
- `git diff --check`

Related: #251
Related: #324

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -142,6 +142,7 @@ dependencies {
     implementation(libs.androidx.compose.ui.graphics)
     implementation(libs.androidx.compose.ui.tooling.preview)
     implementation(libs.androidx.compose.material3)
+    implementation(libs.androidx.compose.material3.adaptive.navigation.suite)
     implementation(libs.androidx.compose.material.icons.extended)
     implementation(libs.coil.compose)
     implementation(libs.coil.network.okhttp)
--- a/app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
+++ b/app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
@@ -15,6 +15,7 @@ import androidx.compose.foundation.selection.selectable
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.BoxWithConstraints
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Spacer
@@ -51,8 +52,12 @@ import androidx.compose.material.icons.rounded.GridView
 import androidx.compose.material.icons.rounded.History
 import androidx.compose.material.icons.rounded.KeyboardArrowDown
 import androidx.compose.material.icons.rounded.KeyboardArrowUp
+import androidx.compose.material.icons.rounded.Person
 import androidx.compose.material.icons.rounded.Settings
 import androidx.compose.material.icons.rounded.Search
+import androidx.compose.material3.adaptive.ExperimentalMaterial3AdaptiveApi
+import androidx.compose.material3.adaptive.navigationsuite.NavigationSuiteScaffold
+import androidx.compose.material3.adaptive.navigationsuite.NavigationSuiteType
 import androidx.compose.material3.Button
 import androidx.compose.material3.FilterChip
 import androidx.compose.material3.Card
@@ -85,6 +90,8 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.input.nestedscroll.NestedScrollConnection
 import androidx.compose.ui.input.nestedscroll.NestedScrollSource
 import androidx.compose.ui.input.nestedscroll.nestedScroll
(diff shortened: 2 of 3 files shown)


app/build.gradle.kts
```diff
@@ -142,6 +142,7 @@ dependencies {
     implementation(libs.androidx.compose.ui.graphics)
     implementation(libs.androidx.compose.ui.tooling.preview)
     implementation(libs.androidx.compose.material3)
+    implementation(libs.androidx.compose.material3.adaptive.navigation.suite)
```
PengDingkang, line 145: Dependency addition looks good — pinned via `version.ref = "material3"` in the version catalog, consistent with the existing `androidx-compose-material3` entry, no open range.

app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
```diff
@@ -757,194 +770,228 @@ private fun BrowsePager(
...
+                        item(
+                            icon = {
+                                Icon(
+                                    imageVector = section.navigationIcon(),
+                                    contentDescription = section.localizedLabel(),
+                                )
+                            },
+                            label = { Text(section.localizedLabel()) },
+                            selected = uiState.selectedBrowseSection == section,
+                            onClick = { onSelectBrowseSection(section) },
```
PengDingkang, line 929: Minor: `containerColor = Color.Transparent` is a one-off hardcoded color for the rail background. Not a big deal here since it's just "no background" rather than a duplicated alpha/radius token, but if there's ever a need for a themed rail surface (e.g., to visually separate it from the `background` brush behind `content()`), consider sourcing it from `SakiTheme.visuals` like the rest of the Browse chrome does, rather than a raw `Color` literal.

app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
```diff
@@ -757,194 +770,228 @@ private fun BrowsePager(
...
+        if (useAdaptiveNavigation) {
+            NavigationSuiteScaffold(
+                navigationSuiteItems = {
+                    sections.forEach { section ->
+                        item(
+                            icon = {
+                                Icon(
+                                    imageVector = section.navigationIcon(),
+                                    contentDescription = section.localizedLabel(),
+                                )
```
PengDingkang, line 925: Each nav item pairs `Icon(contentDescription = section.localizedLabel())` with a visible `Text(section.localizedLabel())` right next to it. TalkBack will announce the label twice for every rail item. The rest of this file (and `BrowseComponents.kt`, `SettingsScreen.kt`, `Artwork.kt`, `PlayerChrome.kt`) consistently uses `contentDescription = null` on icons that already have an adjacent visible label — worth following that same pattern here:

```kotlin
icon = {
    Icon(
        imageVector = section.navigationIcon(),
        contentDescription = null,
    )
},
label = { Text(section.localizedLabel()) },
```

app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
```diff
@@ -757,194 +770,228 @@ private fun BrowsePager(
...
+                                        )
+
+                                        BrowseSection.SONGS -> SongsPageRoute(
+                                            browsePagerState = pagerState,
+                                            songs = uiState.songs,
+                                            songsOffset = uiState.songsOffset,
+                                            hasPrevious = uiState.hasPreviousSongs,
+                                            hasMore = uiState.hasMoreSongs,
+                                            server = currentServer,
+                                            playbackUiStateFlow = playbackUiStateFlow,
```
PengDingkang, line 886: The search overlay (`if (uiState.isSearchActive) { ... }`) lives inside `content()`, which is placed in `NavigationSuiteScaffold`'s content slot when `useAdaptiveNavigation` is true. That means on wide screens the search overlay's `Column.fillMaxSize()` fills only the content pane area next to the rail, not the full window — so the rail stays visible while searching. Is that the intended behavior on tablet/desktop widths, or should search (like Settings/Now Playing) cover the entire window including the rail? Worth confirming against issue #324's "Verify ... Search overlays still layer correctly over the adaptive shell" acceptance criterion.

PengDingkang (commented):
Reviewed against #324 task 1 (root adaptive nav chrome). Overall approach is sound: `BoxWithConstraints` + 600dp threshold matches the M3 medium breakpoint, the compact phone path is untouched (chip row still renders unchanged below 600dp, same `HorizontalPager`/pull-to-refresh/search-overlay structure), and the dependency is pinned correctly through the version catalog. Since this app has no separate root tab bar (Browse's section tabs are effectively the root nav), wrapping `NavigationSuiteScaffold` around `BrowsePager`'s content is a reasonable reading of "root navigation chrome" for this codebase.

Three things worth addressing before merge:
1. Icon `contentDescription` duplicates the adjacent visible label text in each rail item, causing double TalkBack announcements — should be `null` per the pattern used everywhere else in this file and elsewhere in the app.
2. Minor: hardcoded `Color.Transparent` for `containerColor` — fine as-is, just flagging as a spot that diverges from `SakiTheme.visuals`-sourced values if a themed rail background is ever wanted.
3. Need confirmation on whether the search overlay should span the full window (covering the rail) or just the content pane on wide screens — currently it's scoped to the content pane only.

CI build was still in progress at review time; please confirm `assembleDebug`/`testDebugUnitTest` pass before merging.

PengDingkang:
Addressed the actionable review items in `aaab5be`:

- Changed the navigation rail item icon `contentDescription` to `null` so TalkBack uses the adjacent visible label instead of announcing each item twice.
- Moved the search overlay outside the `NavigationSuiteScaffold` content slot so, on wide layouts, search covers the full Browse surface including the rail.

Left `containerColor = Color.Transparent` unchanged for now because this PR intentionally keeps the rail background visually neutral; we can tokenise a themed rail surface later if the large-screen design needs separation.

Validation:
- `git diff --check`

MYmeru (approved):

PR #326, opened 2026-07-02
[codex] Adapt album grid columns

PengDingkang:
## Summary
- Keep compact album grid behavior at two columns.
- Switch album grid to adaptive columns once content width reaches 520dp.
- Use a 168dp adaptive album cell target for landscape, foldable, tablet, and desktop-sized Browse layouts.

## Scope
Implements the album grid portion of #324. This does not change list mode, album cards, fast-scroll behavior, or detail-page layout.

## Validation
- `git diff --check`

Related: #251
Related: #324

### DIFF
--- a/app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
+++ b/app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
@@ -148,6 +148,8 @@ import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.withContext
 
 private val BrowseAdaptiveNavigationMinWidth = 600.dp
+private val AlbumAdaptiveGridMinContentWidth = 520.dp
+private val AlbumAdaptiveGridMinCellWidth = 168.dp
 
 @OptIn(ExperimentalFoundationApi::class)
 @Composable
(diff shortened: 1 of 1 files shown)


PengDingkang (commented):
Small, well-scoped change implementing the album-grid portion of #324. Verified:

- `AlbumCard` (in `BrowseComponents.kt`) has no fixed width — it uses `fillMaxWidth().aspectRatio(1f)` for artwork, so it stretches correctly under both `GridCells.Fixed(2)` and `GridCells.Adaptive(168.dp)`.
- No `horizontalArrangement`/spacing is set on this `LazyVerticalGrid`, so the adaptive column math is straightforward.
- `GridItemSpan(maxLineSpan)` used for loading/error/empty states resolves correctly regardless of which `GridCells` strategy is active, so those states remain full-width under adaptive too.
- The 520dp switch threshold is a sensible choice: it keeps phones on a predictable fixed 2-column layout and only switches to fluid adaptive once there's room for 3+ columns, avoiding an awkward "adaptive but still only 2 columns" zone right at the boundary.
- Change is isolated to `albumGridCells()` + one call site; list mode, album cards, fast-scroll, and detail pages are untouched as stated in the PR scope.

No issues found. CI `build` check was still in progress at review time — please confirm `assembleDebug` passes before merging. (Note: couldn't submit as an "Approve" since this account is the PR author — GitHub blocks self-approval.)

MYmeru (approved):

PR #327, opened 2026-07-02
fix: refine adaptive browse layout

PengDingkang:
## Summary
- Replace the default adaptive navigation rail treatment with a lighter custom rail indicator so the left rail no longer reads as a heavy slab.
- Add a wide-screen album/playlist detail hero layout that keeps artwork bounded and places metadata/actions beside it.
- Remove the now-unused adaptive navigation suite dependency.

## Validation
- Installed on `10.233.233.69:34567` and checked screenshots for Browse rail, album grid, album detail, and playlist detail.

Refs #324

### DIFF
--- a/app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
+++ b/app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
@@ -22,6 +22,7 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.size
@@ -55,9 +56,6 @@ import androidx.compose.material.icons.rounded.KeyboardArrowUp
 import androidx.compose.material.icons.rounded.Person
 import androidx.compose.material.icons.rounded.Settings
 import androidx.compose.material.icons.rounded.Search
-import androidx.compose.material3.adaptive.ExperimentalMaterial3AdaptiveApi
-import androidx.compose.material3.adaptive.navigationsuite.NavigationSuiteScaffold
-import androidx.compose.material3.adaptive.navigationsuite.NavigationSuiteType
 import androidx.compose.material3.Button
 import androidx.compose.material3.FilterChip
 import androidx.compose.material3.Card
@@ -148,6 +146,12 @@ import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.withContext
 
 private val BrowseAdaptiveNavigationMinWidth = 600.dp
+private val BrowseAdaptiveNavigationRailWidth = 104.dp
+private val BrowseAdaptiveNavigationRailItemWidth = 88.dp
+private val BrowseAdaptiveNavigationRailItemHeight = 76.dp
+private val BrowseAdaptiveNavigationRailIndicatorWidth = 64.dp
+private val BrowseAdaptiveNavigationRailIndicatorHeight = 36.dp
+private val BrowseAdaptiveNavigationContentGap = 12.dp
 private val AlbumAdaptiveGridMinContentWidth = 520.dp
 private val AlbumAdaptiveGridMinCellWidth = 168.dp
 
@@ -711,7 +715,6 @@ private fun OfflineModeBanner(modifier: Modifier = Modifier) {
 
 @OptIn(
     ExperimentalFoundationApi::class,
-    ExperimentalMaterial3AdaptiveApi::class,
     ExperimentalMaterial3ExpressiveApi::class,
 )
 @Composable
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
```diff
@@ -998,6 +997,75 @@ private fun BrowsePager(
...
+private fun BrowseAdaptiveNavigationRail(
+    sections: List<BrowseSection>,
+    selectedSection: BrowseSection,
+    onSelectBrowseSection: (BrowseSection) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    Column(
+        modifier = modifier.padding(top = 8.dp),
+        horizontalAlignment = Alignment.CenterHorizontally,
+        verticalArrangement = Arrangement.spacedBy(8.dp),
```
PengDingkang, line 1010: This custom rail item uses a plain `Surface(onClick = ...)`. The `NavigationSuiteScaffold` item this replaces exposed selected/Tab semantics to accessibility services automatically, and the sibling `BrowseSectionChip`/`BrowseFeedChip` in this same file use `Modifier.selectable(selected = selected, onClick = onClick, role = Role.Tab)` for exactly this reason (TalkBack announces "Tab, selected"/"Tab, not selected"). This rail item doesn't carry that state at all now — TalkBack will just announce it as a generic clickable with no selected/tab context.

Suggest matching the existing convention:
```kotlin
Surface(
    modifier = Modifier
        .width(BrowseAdaptiveNavigationRailItemWidth)
        .height(BrowseAdaptiveNavigationRailItemHeight)
        .selectable(
            selected = selected,
            onClick = { onSelectBrowseSection(section) },
            role = Role.Tab,
        ),
    shape = MaterialTheme.shapes.large,
    color = Color.Transparent,
    contentColor = MaterialTheme.colorScheme.onSurfaceVariant,
) { ... }
```
(dropping the `onClick` param from `Surface` itself since `selectable` already provides the click handling).

app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
```diff
@@ -148,6 +146,12 @@ import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.withContext
 
 private val BrowseAdaptiveNavigationMinWidth = 600.dp
+private val BrowseAdaptiveNavigationRailWidth = 104.dp
+private val BrowseAdaptiveNavigationRailItemWidth = 88.dp
```
PengDingkang, line 150: Good cleanup: removing the now-unused `material3-adaptive-navigation-suite` dependency from both `build.gradle.kts` and the version catalog, and dropping the now-unused `ExperimentalMaterial3AdaptiveApi` opt-in. `WideAlbumDetailHeroCard` reuses the existing `ensureContrast`/`contrastRatio` helpers consistently with `CompactAlbumDetailHeroCard`, and playlist detail picks up the wide layout for free since it already routes through the shared `AlbumDetailHeroCard`.

app/src/main/java/org/hdhmc/saki/presentation/library/BrowseScreen.kt
```diff
@@ -998,6 +997,75 @@ private fun BrowsePager(
...
+                            Icon(
+                                imageVector = section.navigationIcon(),
+                                contentDescription = null,
+                                modifier = Modifier.size(26.dp),
+                            )
+                        }
+                    }
+                    Spacer(modifier = Modifier.height(4.dp))
+                    Text(
+                        text = section.localizedLabel(),
```
PengDingkang, line 1058: Minor: `BrowseAdaptiveNavigationRailWidth` (104dp) is a fixed width regardless of label length across locales. `section.localizedLabel()` is translated (see `strings.xml`/i18n docs referenced in project conventions) — worth a quick check that longer translations (e.g., German, or CJK with `maxLines = 1`) don't get clipped awkwardly within the 88dp item width. Not blocking since `maxLines = 1` on the label `Text` will just ellipsis-truncate... actually there's no `overflow = TextOverflow.Ellipsis` set here, so a long label will likely just get cut off by layout rather than showing an ellipsis. Consider adding `overflow = TextOverflow.Ellipsis` to the label `Text` for clarity in longer locales.

PengDingkang (commented):
Reviewed against the wider adaptive-navigation work in #324/#325/#326. This PR trades the stock `NavigationSuiteScaffold` rail for a lighter custom one and adds a wide hero layout for album/playlist detail — both reasonable visual refinements once the mechanism from #325 was in place.

One accessibility regression worth fixing before merge: the new custom rail item (`BrowseAdaptiveNavigationRail`) drops the `selected`/`Role.Tab` semantics that both the removed `NavigationSuiteScaffold` item and the sibling `BrowseSectionChip`/`BrowseFeedChip` pattern in this same file provide via `Modifier.selectable(...)`. TalkBack will no longer announce selection state for rail items. See inline comment for the fix (swap `Surface(onClick = ...)` for `Surface(modifier = Modifier...selectable(selected = selected, onClick = ..., role = Role.Tab))`).

Also flagged a minor i18n robustness note on the rail label `Text` missing `TextOverflow.Ellipsis` for longer translated labels — non-blocking.

Positives: clean removal of the now-unused adaptive-navigation-suite dependency from both `build.gradle.kts` and the version catalog (plus the now-unnecessary `ExperimentalMaterial3AdaptiveApi` opt-in), and the wide hero layout is shared correctly between album and playlist detail via the existing `AlbumDetailHeroCard` entry point, matching the PR's stated scope. `WideAlbumDetailHeroCard` reuses the established `ensureContrast`/`contrastRatio` contrast-safety helpers and `minimumInteractiveComponentSize()` convention consistently with the rest of the file.

CI `build` check was still in progress at review time — please confirm `assembleDebug` passes, and ideally do a quick TalkBack pass on the rail after the semantics fix.

PengDingkang:
Addressed the review feedback in `254139e`:

- Restored rail item accessibility semantics by replacing `Surface(onClick = ...)` with `Modifier.selectable(selected = ..., role = Role.Tab)`.
- Added `TextOverflow.Ellipsis` to the rail label for longer localized labels.

Validation: `./gradlew assembleDebug -q` and `git diff --check` pass locally.

MYmeru (approved):

PR #328, opened 2026-07-02
[codex] Improve landscape mini player layout

PengDingkang:
## Summary
- Constrain the mini player capsule width on compact-height landscape screens so it behaves like a centered floating surface instead of a full-width bottom bar.
- Compact wide album/playlist detail heroes in compact landscape to keep the hero content visible above the mini player overlay.

## Validation
- Installed debug build on `10.111.111.90:34567`
- Captured and reviewed landscape screenshots for songs, albums list/grid, album detail, playlist detail, settings, search, and Now Playing.

Refs #324

### DIFF
--- a/app/src/main/java/org/hdhmc/saki/presentation/library/BrowseComponents.kt
+++ b/app/src/main/java/org/hdhmc/saki/presentation/library/BrowseComponents.kt
@@ -78,6 +78,7 @@ import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.graphics.Brush
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.luminance
+import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.res.pluralStringResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.font.FontWeight
@@ -104,6 +105,8 @@ import org.hdhmc.saki.ui.theme.SakiTheme
 
 private val LibraryDetailWideHeroMinWidth = 720.dp
 private val LibraryDetailWideHeroArtworkWidth = 320.dp
+private val LibraryDetailWideHeroCompactArtworkWidth = 220.dp
+private const val LibraryDetailWideHeroCompactMaxHeightDp = 520
 
 @Composable
 fun ArtistDetailScreen(
@@ -506,11 +509,21 @@ private fun AlbumDetailHeroCard(
     onBack: () -> Unit,
     accentColor: Color,
 ) {
+    val configuration = LocalConfiguration.current
+    val useCompactLandscapeHero = configuration.screenWidthDp > configuration.screenHeightDp &&
+        configuration.screenHeightDp < LibraryDetailWideHeroCompactMaxHeightDp
+    val wideHeroArtworkWidth = if (useCompactLandscapeHero) {
+        LibraryDetailWideHeroCompactArtworkWidth
+    } else {
+        LibraryDetailWideHeroArtworkWidth
+    }
+
     BoxWithConstraints(modifier = Modifier.fillMaxWidth()) {
         if (maxWidth >= LibraryDetailWideHeroMinWidth) {
             WideAlbumDetailHeroCard(
                 title = title,
                 artwork = artwork,
+                artworkWidth = wideHeroArtworkWidth,
                 metaItems = metaItems,
                 canPlay = canPlay,
                 onPlay = onPlay,
@@ -535,6 +548,7 @@ private fun AlbumDetailHeroCard(
 private fun WideAlbumDetailHeroCard(
     title: String,
     artwork: Any?,
+    artworkWidth: Dp,
     metaItems: List<String>,
     canPlay: Boolean,
     onPlay: () -> Unit,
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/hdhmc/saki/presentation/library/PlayerChrome.kt
```diff
@@ -224,150 +225,166 @@ fun NowPlayingCapsule(
...
-                        didOpen = true
-                        onExpandState.value()
+        Card(
+            onClick = onExpand,
+            enabled = track != null,
+            modifier = Modifier
+                .then(
+                    if (constrainLandscapeCapsuleWidth) {
+                        Modifier.widthIn(max = MINI_PLAYER_LANDSCAPE_MAX_WIDTH)
+                    } else {
```
PengDingkang, line 244: Good use of `BoxWithConstraints` + `maxWidth`/`maxHeight` here to detect the compact-landscape case — this is measured against the actual available space for the capsule, correctly handling multi-window/split-screen. This is the more robust approach compared to the `LocalConfiguration`-based check added in `BrowseComponents.kt` (see comment there). The `widthIn(max = MINI_PLAYER_LANDSCAPE_MAX_WIDTH).fillMaxWidth()` ordering is correct — the outer `widthIn` caps the constraint before `fillMaxWidth()` fills within it. The existing swipe-to-expand `pointerInput` remains the last modifier in the chain, same relative position to `Card`'s `onClick` as before this refactor, so gesture behavior should be unaffected.

app/src/main/java/org/hdhmc/saki/presentation/library/BrowseComponents.kt
```diff
@@ -506,11 +509,21 @@ private fun AlbumDetailHeroCard(
     onBack: () -> Unit,
     accentColor: Color,
 ) {
+    val configuration = LocalConfiguration.current
+    val useCompactLandscapeHero = configuration.screenWidthDp > configuration.screenHeightDp &&
+        configuration.screenHeightDp < LibraryDetailWideHeroCompactMaxHeightDp
+    val wideHeroArtworkWidth = if (useCompactLandscapeHero) {
+        LibraryDetailWideHeroCompactArtworkWidth
+    } else {
```
PengDingkang: This uses `LocalConfiguration.current.screenWidthDp`/`screenHeightDp`, which reflect the whole device screen, not the actual space available to this composable. In split-screen/multi-window (or a foldable in a non-standard posture), the app's window can be narrower/shorter than the full screen, so this can compute `useCompactLandscapeHero` incorrectly — e.g., the app is in a narrow split-screen pane while the device itself is still "landscape" with a tall `screenHeightDp`.

Right below this, `BoxWithConstraints` + `maxWidth >= LibraryDetailWideHeroMinWidth` correctly measures the actual available constraints for this composable rather than the whole screen. The `NowPlayingCapsule` change in this same PR does it the right way too (`BoxWithConstraints` + `maxWidth`/`maxHeight`). Suggest computing `useCompactLandscapeHero` from the same `BoxWithConstraints` scope instead of `LocalConfiguration`, for consistency and correctness under multi-window:

```kotlin
BoxWithConstraints(modifier = Modifier.fillMaxWidth()) {
    val useCompactLandscapeHero = maxWidth > maxHeight && maxHeight < LibraryDetailWideHeroCompactMaxHeight
    val wideHeroArtworkWidth = if (useCompactLandscapeHero) { ... } else { ... }
    if (maxWidth >= LibraryDetailWideHeroMinWidth) { ... }
}
```
(would need `LibraryDetailWideHeroCompactMaxHeightDp` converted to a `Dp` constant to compare against `maxHeight` directly, mirroring `MINI_PLAYER_LANDSCAPE_WIDTH_LIMIT_HEIGHT` in `PlayerChrome.kt`.)

PengDingkang (commented):
Reviewed the compact-height landscape refinements for the mini player capsule and detail hero, continuing the #324 adaptive work. Both changes address real layout problems (full-width bottom bar looking wrong in short landscape windows; wide hero artwork eating too much vertical space when height is constrained) with reasonable fixed thresholds (480dp/520dp).

One correctness concern: `AlbumDetailHeroCard`'s new `useCompactLandscapeHero` check uses `LocalConfiguration.current.screenWidthDp/screenHeightDp`, which reflects the whole device screen rather than the actual space available to the composable. This can misfire under split-screen/multi-window, where the app's window is smaller than the full screen. The `BoxWithConstraints`-based check a few lines below in the same function (and the equivalent check added to `NowPlayingCapsule` in this same PR) does this correctly by measuring actual layout constraints. Suggest moving the compact-landscape check into the same `BoxWithConstraints` scope for consistency — see inline comment for the specific change.

Everything else looks solid: modifier ordering for the capsule's `widthIn(max=...).fillMaxWidth()` is correct, and the swipe-to-expand `pointerInput` retains its original relative position in the modifier chain so gesture behavior should carry over unchanged.

CI `build` check was still in progress at review time.

app/src/main/java/org/hdhmc/saki/presentation/library/BrowseComponents.kt
PengDingkang: Fixed in `b36be9b`: moved the compact-landscape hero check into the existing `BoxWithConstraints` scope and compare against a `Dp` height threshold using the actual composable constraints, matching the mini player approach.

MYmeru (approved):

PR #330, opened 2026-07-10
fix: align Now Playing auto-scroll behavior

PengDingkang:
## Summary
- replace the track title's one-way `basicMarquee` with the same edge-paused, bidirectional auto-scroll behavior used by artist and album metadata
- share overflow detection, motion-scale handling, velocity, pauses, and reset behavior between both rows while keeping independent scroll distances
- retain the static two-line title treatment on large screens

## Root cause
The title used Compose's repeating marquee while the adjacent clickable metadata row used a custom ping-pong `horizontalScroll`. Their different direction and loop behavior made long Now Playing metadata look inconsistent. The clickable metadata cannot safely use duplicated marquee content, so both rows now share the non-duplicating scroll controller instead.

## Validation
- `./gradlew testDebugUnitTest assembleDebug -q`
- `./gradlew assembleRelease --console=plain --max-workers=2`
- release APK installed on `10.111.111.90:34567`; Now Playing smoke test completed
- `git diff --check`

### DIFF
--- a/app/src/main/java/org/hdhmc/saki/presentation/library/PlayerChrome.kt
+++ b/app/src/main/java/org/hdhmc/saki/presentation/library/PlayerChrome.kt
@@ -2186,13 +2186,22 @@ private fun PressScaleIconButton(
 
 
 @Composable
-private fun rememberNowPlayingAutoScrollState(scrollKey: Any?): ScrollState {
+private fun currentMotionDurationScaleFactor(): Float {
+    val compositionScope = rememberCoroutineScope()
+    return compositionScope.coroutineContext[MotionDurationScale.Key]?.scaleFactor ?: 1f
+}
+
+@Composable
+private fun rememberNowPlayingAutoScrollState(
+    scrollKey: Any?,
+    motionDurationScaleFactor: Float,
+): ScrollState {
     val scrollState = rememberScrollState()
     val density = LocalDensity.current
-    LaunchedEffect(scrollKey, scrollState.maxValue, density) {
+    LaunchedEffect(scrollKey, scrollState.maxValue, density, motionDurationScaleFactor) {
         scrollState.scrollTo(0)
         if (scrollState.maxValue <= 0 || scrollState.maxValue == Int.MAX_VALUE) return@LaunchedEffect
-        if (coroutineContext[MotionDurationScale.Key]?.scaleFactor == 0f) return@LaunchedEffect
+        if (!nowPlayingAutoScrollEnabled(motionDurationScaleFactor)) return@LaunchedEffect
 
         delay(NOW_PLAYING_AUTO_SCROLL_EDGE_PAUSE_MS)
         val speedPxPerMs = with(density) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/hdhmc/saki/presentation/library/PlayerChrome.kt
```diff
@@ -2182,6 +2185,72 @@ private fun PressScaleIconButton(
...
 
 
+@Composable
+private fun rememberNowPlayingAutoScrollState(scrollKey: Any?): ScrollState {
+    val scrollState = rememberScrollState()
+    val density = LocalDensity.current
+    LaunchedEffect(scrollKey, scrollState.maxValue, density) {
+        scrollState.scrollTo(0)
+        if (scrollState.maxValue <= 0 || scrollState.maxValue == Int.MAX_VALUE) return@LaunchedEffect
+        if (coroutineContext[MotionDurationScale.Key]?.scaleFactor == 0f) return@LaunchedEffect
```
PengDingkang: **[P2] 请为关闭系统动画的情况提供可读的静态 fallback。** 当 `MotionDurationScale` 为 0 时，这里把位置重置为 0 后直接退出；但 `NowPlayingAutoScrollingText` 仍把标题放在 `horizontalScroll(enabled = false)` 的无限宽子项中，并使用 `TextOverflow.Clip`。因此所有溢出的非大屏标题都会永久只显示前缀，既没有省略号，也无法手动查看剩余内容。另一个同根问题是 scale 只在 effect 启动时采样：页面保持组合时从 1x 切到 0，循环可能每隔约 1.2 秒在两端瞬移；从 0 切回 1x 也不会重新启动。建议把 scale 作为可观察状态，在 0 时渲染受 viewport 宽度约束的静态 `Text(maxLines = 1, overflow = Ellipsis)`，并在 scale 变化时取消、归零及按需重启滚动。

PengDingkang (commented):
审查基于 head `5725f04`。发现 1 个系统动画关闭路径的可读性与生命周期回归，详见 inline comment。

app/src/main/java/org/hdhmc/saki/presentation/library/PlayerChrome.kt
PengDingkang: 已在 `0ada52b` 修复。现在通过 composition-owned coroutine scope 读取 `MotionDurationScale.scaleFactor`；该值由 Compose `MutableFloatState` 驱动，因此运行时切换动画倍率会触发重组并重启对应 effect。切到 `0x` 时会取消正在进行的滚动、将元数据位置归零，并把超长标题切换为受 viewport 约束的单行 `TextOverflow.Ellipsis` 静态展示；恢复到非零倍率后从起点重新执行停顿/往返滚动，不再出现延迟瞬移。补充了倍率启停回归测试，`testDebugUnitTest`、`assembleDebug`、`assembleRelease` 和 `git diff --check` 均通过。

MYmeru (approved):

